PR #11, opened 2016-05-04
Show list of XHR request URLs in popup

### LINKED ISSUE #4: Show the list of detected XHR requests in the popup
When interceptor is enabled for a browser tab, all XHR requests should be logged and listed in the popup UI

emilsoman:
Closes #4

### DIFF
--- a/app/background.js
+++ b/app/background.js
@@ -1,7 +1,10 @@
+import MessageService from './message_service'
+
 class BackgroundWorker {
 
   constructor() {
     this.data = {};
+    this.messageService = new MessageService();
   }
 
   startMessageListener() {
@@ -36,6 +39,7 @@ class BackgroundWorker {
       (details) => {
         this.data[tabId].count += 1;
         chrome.browserAction.setBadgeText({text: '' + this.data[tabId].count, tabId: tabId});
+        this.messageService.logRequest(tabId, details);
       },
       {urls: ["<all_urls>"], types: ["xmlhttprequest"], tabId: tabId},
       ["blocking"]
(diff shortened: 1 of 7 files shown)


app/content_script.js
```diff
@@ -0,0 +1,9 @@
+import { requestStore } from './store'
+
+chrome.runtime.onMessage.addListener((request, sender, sendResponse) => {
+  if(request.message == "LOG_REQUEST"){
```
kgrz: is `===` better here?

app/background.js
```diff
@@ -36,6 +39,7 @@ class BackgroundWorker {
       (details) => {
         this.data[tabId].count += 1;
         chrome.browserAction.setBadgeText({text: '' + this.data[tabId].count, tabId: tabId});
```
kgrz: This is not modified in this PR, but interpolation would be better IMO. Babel has support afaik

emilsoman: What I'm doing here is convert an integer into a string using the old trick `'' + integer` . Do you mean to say do this instead? : `` `${integer}` ``

kgrz: yeah, that's what I meant

emilsoman:
Closing this because I thought of a better way to show URLs . See #12 .

Problem with this PR is the following:

A popup's lifespan starts when you click on the extension's browser button and stops when you close it.
This means you cannot persist stuff in the markup. Another problem is that the popup cannot listen to messages from the backend worker script.

Moving this to a content script makes sense because a content script will be alive as long as the tab is open, and it can listen to events from the backend worker - which means we can show a live updating list of XHR urls.

PR #21, opened 2017-05-02
Try out Typescript

kgrz:


webpack.config.js
```diff
@@ -1,28 +1,34 @@
...
-        test: /\.js$/,
-        loaders: ["react-hot", "babel"],
-        include: path.join(__dirname, 'app')
+        test: /\.tsx?$/,
+        loader: "awesome-typescript-loader"
       },
       { test: /\.css$/,
-        loader: "style-loader!css-loader",
-        include: path.join(__dirname, 'app')
+        loader: "style-loader!css-loader"
```
kgrz: nope, I just kept it thinking we might get around to doing it properly next time.

webpack.config.js
```diff
@@ -31,6 +37,11 @@ module.exports = {
         'NODE_ENV': `"${process.env.NODE_ENV}"`
       }
     }),
-    new CommonsChunkPlugin("common.js", ["popup", "content_script"])
+    new CommonsChunkPlugin("common.js", ["popup", "content_script"]),
+    new CopyWebpackPlugin([
+      { from: 'manifest.json' },
+      { from: 'index.html' },
+      { from: 'styles.css' }
```
kgrz: I thought about it, but I generally saw dist being not gitignored, so I didn't remove it. Will make it so.

tsconfig.json
```diff
@@ -0,0 +1,24 @@
...
+		"pretty": true,
+		"removeComments": true,
+		"strictNullChecks": true
+	},
+	"include": [
+		"./app/**/*"
+	],
+	"exclude": [
+		"node_modules"
+	]
```
kgrz: Good catch, will remove this

dist/manifest.json
```diff
@@ -18,13 +18,6 @@
     "webRequestBlocking",
     "http://*/*",
     "https://*/*"
-  ],
-
-  "content_scripts": [
-    {
-      "matches": ["<all_urls>"],
-      "js": ["common.js", "content_script.js"]
-    }
```
kgrz: this is needed only if we are using content script functionality, which at this point we are not. So I removed it for some clarity. will add it back once we decide how to show the recorded log. I'm thinking a new tab is better. We can discuss this with some prototypes.

app/watch.tsx
```diff
@@ -0,0 +1,30 @@
...
+interface WatchProps {
+
+}
+
+interface WatchState {
+  watchRegex: string
+}
+
+class Watch extends React.Component<WatchProps, WatchState> {
+  state = { watchRegex: "*" }
```
kgrz: For now I have so many places where I might've forgotten adding semicolons. will add that to a linter and fix all at once out of this PR.

dist/styles.css
```diff
@@ -0,0 +1,52 @@
...
+  cursor: pointer;
+}
+
+.button-start-listening {
+  background: #03A9F4;
+}
+
+.button-stop-listening {
+  background: #8bc34a;
+}
```
kgrz: Will do.

app/background.ts
```diff
@@ -0,0 +1,90 @@
...
+  data : Array<Data> = [];
+
+  findItem = (tabId : number) => {
+    for (let index = 0; index < this.data.length; index++) {
+      const data = this.data[index];
+
+      if (Boolean(data) && data.tabId === tabId) {
+        return data;
+      }
+    }
```
kgrz: I'm not getting find's type annotations in TS, so did this. Will move that to `find` once I figure out that part.

app/background.ts
```diff
@@ -0,0 +1,90 @@
...
+
+    // Lazy code for object cloning
+    return JSON.parse(JSON.stringify(DEFAULT_DATA));
+  }
+
+  startMessageListener() {
+    chrome.runtime.onMessage.addListener((request, _, sendResponse) => {
+      const tabId = request.tabId;
+      const data = this.findItem(tabId);
```
kgrz: True. We might want this for the other branches too, since a tab id is pretty much a requirement for getting tab related info. Anyway, will remove this.

app/content_script.css
```diff
@@ -1,4 +1,4 @@
-#interceptor-container {
+.interceptor-container {
   position: absolute;
   top: 10px;
   right: 10px;
```
kgrz: I think this one is being used in the content script. Since I didn't remove the code, I'm keeping this as it is

tsconfig.json
kgrz: On second thought, I don't get the impression that these are mutually exclusive: https://www.typescriptlang.org/docs/handbook/tsconfig-json.html#details It looks like adding something in `include` doesn't necessary mean only that folder tree is compiled. So we might have to add the `exclude` directive as well. 🤔

webpack.config.js
kgrz: removed `/dist`

tsconfig.json
emilsoman: It says "Files included using "include" can be filtered using the "exclude" property. ". Also looks like "node_modules" is excluded by default. Does it work when you remove the "exclude" part?

dist/manifest.json
emilsoman: 👍

app/background.ts
emilsoman: Cool. Please add a TODO comment so we can do it later.

app/content_script.css
emilsoman: Since content script is removed, we might as well remove everything used by the content script.

tsconfig.json
kgrz: It ran fine without any noticeable delay. Removed it.

PR #22, opened 2018-01-16
List requests to view

amit-mb:
The requests urls and the request method are displayed now. Used MessageService to get the incoming requests and added the requests to state in Popup component which is passed as props to RequestList component which was already mapping the urls to a table. 

As of now, only the request url's and methods are rendered.


app/background.ts
```diff
@@ -15,7 +15,7 @@ const DEFAULT_DATA : Data = {
   count: 0
 };
 
-class BackgroundWorker {
+export class BackgroundWorker {
```
revathskumar: Why we need to export this?

app/background.ts
```diff
@@ -31,6 +31,8 @@ class BackgroundWorker {
     chrome.runtime.onMessage.addListener((request, _, sendResponse) => {
       const tabId = request.tabId;
 
+      //console.log('message service :: ', request);
```
revathskumar: remove log.

app/getRequests.tsx
```diff
@@ -0,0 +1,11 @@
+let requestsArr:Array<Object> = [{url : "", requestBody: undefined , method : "", type: ""}]
```
revathskumar: This is not using anywhere. so we can remove this.

app/popup.tsx
```diff
@@ -11,18 +13,26 @@ const isChromeUrl = (url : string) => {
...
       label: props.enabled ? 'Stop Listening' : 'Start Listening',
-      errorMessage: ''
+      errorMessage: '',
+      requests: [],
     };
   }
 
+  componentDidMount() {
+    MessageService.getRequests(this.props.tabId, (requests) => {
+      console.log('reqs :: ', requests);
```
revathskumar: can remove the logs

app/getRequests.tsx
```diff
@@ -0,0 +1,11 @@
...
+
+chrome.webRequest.onBeforeRequest.addListener(
+function(details) {
```
kgrz: Try spend some time setting up your editor. And try using https://github.com/prettier/prettier

amit-mb: Sure Kashyap. Will do that and follow .editorconfig

app/popup.tsx
amit-mb: Will do the changes and push accordingly

package-lock.json
```diff
@@ -0,0 +1,3487 @@
+{
```
revathskumar: please remove the package-lock.json. we already have yarn.lock

PR #24, opened 2018-01-19
Fixes #23

### LINKED ISSUE #23: Ignore package-lock.json
Since we already have `yarn.lock`. Ignore `package-lock.json` and avoid generating this using a project level .npmrc.

amit-mb:


package.json
```diff
@@ -17,6 +17,7 @@
     "build": "./build",
     "watch": "webpack --progress --colors --watch"
   },
+  "lock": false,
```
revathskumar: @amit-mb Can you point me to the documentation for this?

revathskumar:
@amit-mb Can have better commit message saying `ignore package-lock.json` or something like that. 😉

package.json
amit-mb: Sure, https://docs.npmjs.com/misc/config#package-lock

amit-mb: Actually referred this, https://github.com/npm/npm/issues/16867#issuecomment-324554854

revathskumar: Thats is just a proposal by someone. I don't think npm have any plan for this to implement.

emilsoman:
Yeah `Fixes #...` should go into the description. It works the same way of closing the issue when merged.

PR #28, opened 2018-01-23
Log request headers

amit-mb:
Logs requests on a per tab basis. Also fixes the button states across tabs.


app/manifest.json
```diff
@@ -5,13 +5,12 @@
   "version": "0.1",
 
   "background": {
-    "scripts": ["background.js"]
+    "scripts": ["background.js"],
+    "persistent": true
```
revathskumar: why we need this?

app/background.ts
```diff
@@ -8,83 +8,84 @@ interface Data {
...
-  findItem = (tabId : number) => {
-    // Not using a polyfill because this is supported in Chrome v. > 45. We
-    // might have to add some restriction in manifest related to this
-    const data = this.data.find(item => item.tabId === tabId) || DEFAULT_DATA;
-
-    // Lazy code for object cloning
-    return JSON.parse(JSON.stringify(data));
+class BackgroundWorker {
+  currentTab:number = -1;
+  data = {}
```
revathskumar: don't we need to specify type here?

app/message_service.ts
```diff
@@ -18,17 +18,12 @@ export function resetData(tabId: number) {
   chrome.tabs.sendMessage(tabId, {message: "RESET_DATA"});
 }
 
-export function getEnabledStatus(callback: GenericCallback) {
-  chrome.runtime.sendMessage({message: "GET_ENABLED_STATUS"}, callback);
+export function getEnabledStatus(tabId:number, callback: GenericCallback) {
+  chrome.runtime.sendMessage({message: "GET_ENABLED_STATUS", tabId: tabId }, callback);
```
revathskumar: instead of `{ tabId: tabId }` we can use `{ tabId }`

app/message_service.ts
```diff
@@ -18,17 +18,12 @@ export function resetData(tabId: number) {
...
 }
 
-export function getRequests(tabId, callback: GenericCallback) {
-  chrome.runtime.sendMessage(chrome.runtime.id, {message: "GET_REQUESTS", tabId: tabId }, {}, callback);
+export function getRequests(tabId:number, callback: GenericCallback) {
+  chrome.runtime.sendMessage({message: "GET_REQUESTS", tabId: tabId }, {}, callback);
```
revathskumar: instead of `{ tabId: tabId }` we can use `{ tabId }`

app/manifest.json
amit-mb: ```"persistent": true``` is by default. We can revert that back

PR #29, opened 2018-01-24
Display badge count

amit-mb:
Displays the number of listened requests as Badge text

### DIFF
--- a/app/background.ts
+++ b/app/background.ts
@@ -1,15 +1,14 @@
 /// <reference path="../node_modules/@types/chrome/chrome-app.d.ts" />
 import { RequestObj } from './request_list'
 
-interface Data {
+interface TabRecord {
   tabId: number,
   enabled: boolean,
   requests: Array<RequestObj>
-  count: number
 };
 
-interface data {
-  [index: number]: Data;
+interface Recordings {
+  [index: number]: TabRecord;
 }
 
 class BackgroundWorker {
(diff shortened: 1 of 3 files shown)


app/background.ts
```diff
@@ -59,6 +67,7 @@ class BackgroundWorker {
     const tabRecords = this.data[this.currentTab];
     if (tabRecords.enabled && this.currentTab === details.tabId) {
       tabRecords.requests.push(details);
+      tabRecords.count += 1
```
revathskumar: why keep this separate? just take the `tabRecords.requests.length`.

app/popup.tsx
```diff
@@ -103,6 +103,13 @@ chrome.tabs.query(queryParams, tabs => {
   const { id, url } = tab;
   if (typeof id === "undefined" || typeof url === "undefined") return;
 
+  MessageService.getCount(id, (count) => {
+  chrome.browserAction.setBadgeText({
+      text: `${count}`,
```
revathskumar: no need for the string interpolation.
just use `text: count`

app/background.ts
```diff
@@ -51,6 +55,10 @@ class BackgroundWorker {
           sendResponse(this.data[this.currentTab].requests);
           break;
         }
+        case 'GET_COUNT' : {
+          sendResponse(this.data[this.currentTab].count);
```
revathskumar: `this.data[this.currentTab].requests.length`

app/background.ts
```diff
@@ -23,6 +23,10 @@ class BackgroundWorker {
   }
 
   startMessageListener() {
+    chrome.browserAction.setBadgeText({
+      text: `${this.data[this.currentTab].count}`,
```
revathskumar: No need for string interpolation.

app/background.ts
amit-mb: Yeah, thats better

app/popup.tsx
amit-mb: We have declared count to be of number type https://github.com/code-mancers/interceptor/blob/4cda610db7930ff9c4042e055adc5799615f118c/app/background.ts#L8 , we need interpolation I guess.

PR #31, opened 2018-01-29
Add redux store

amit-mb:
Added redux store to manage various states and a clear button which clear the ajax request list off the view


app/popup.tsx
```diff
@@ -1,91 +1,79 @@
...
-    const errorMessage = this.state.errorMessage;
-    const isListening = this.state.enabled;
-    const buttonClass = cx("button", {
-      "button-start-listening": !isListening,
-      "button-stop-listening": isListening
-    });
+    const props = this.props
+    const { requests ,enabled, errorMessage } = props;
+
+    const isListening = enabled;
```
revathskumar: `isListening` is not necessary. we can use `enabled` directly.

app/popup.tsx
```diff
@@ -1,91 +1,79 @@
...
+
   render() {
-    const errorMessage = this.state.errorMessage;
-    const isListening = this.state.enabled;
-    const buttonClass = cx("button", {
-      "button-start-listening": !isListening,
-      "button-stop-listening": isListening
-    });
+    const props = this.props
+    const { requests ,enabled, errorMessage } = props;
```
revathskumar: use `this.props` directly here.

app/modules/recordings.tsx
```diff
@@ -0,0 +1,35 @@
+import {PopUpState, Action} from '../types'
```
revathskumar: This file can be just `.ts` instead of `.tsx`

app/modules/recordings.tsx
```diff
@@ -0,0 +1,35 @@
...
+  }
+}
+const clearRequestsArray = (state:PopUpState) => {
+  const clonedArray = state.requests.requests.slice(0)
+  return clonedArray
+}
+const reloadDataReducer = (state:PopUpState, action:Action) => {
+  let incomingRequests:any = [];
+  incomingRequests = action.requests;
+  return {...state, requests : incomingRequests}
```
revathskumar: why unnecessary declare vars?  use `action.requests` directly.

app/modules/recordings.tsx
```diff
@@ -0,0 +1,35 @@
...
+    case actionType.START_LISTENING:
+      return {...state, enabled : true}
+    case actionType.UPDATE_FIELD:
+      return {...state, requests : action.requests };
+    case actionType.STOP_LISTENING: 
+      return {...state, enabled: false};
+    case actionType.ERROR: 
+      return {...state, errorMessage : action.errorMessage }
+    case actionType.CLEAR_REQUESTS : return {...state, requests: clearRequestsArray (state) }
+    case actionType.RELOAD_REQUESTS : reloadDataReducer (state, action)
```
revathskumar: No need for another function here. can do the operation directly

app/modules/recordings.tsx
```diff
@@ -0,0 +1,35 @@
...
+  switch (action.type) {
+    case actionType.START_LISTENING:
+      return {...state, enabled : true}
+    case actionType.UPDATE_FIELD:
+      return {...state, requests : action.requests };
+    case actionType.STOP_LISTENING: 
+      return {...state, enabled: false};
+    case actionType.ERROR: 
+      return {...state, errorMessage : action.errorMessage }
+    case actionType.CLEAR_REQUESTS : return {...state, requests: clearRequestsArray (state) }
```
revathskumar: why cloneArray when clear. set to empty array.

app/actions/index.tsx
```diff
@@ -0,0 +1,29 @@
+import {PopUpState} from '../types'
```
revathskumar: file extension can be `.ts` instead of `.tsx`

app/background.ts
```diff
@@ -47,6 +47,7 @@ class BackgroundWorker {
         }
         case 'GET_REQUESTS': {
           sendResponse(this.data[this.currentTab].requests);
+          console.log(this.data[this.currentTab].requests);
```
revathskumar: remove log

app/actions/index.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+export function updateField (requests:PopUpState["requests"]) {
+  return {type : UPDATE_FIELD, requests}
+}
+
+export function stopListening (enabledStatus:PopUpState["enabled"]){ 
+  return { type : STOP_LISTENING, enabled: enabledStatus}
+ }
+export function errorNotify (tabUrl:PopUpState["tabUrl"]) {
+  return { type : ERROR ,  tabUrl  }
```
revathskumar: in reducer it seems like it expects `errorMessage` not `tabUrl`

app/background.ts
```diff
@@ -86,6 +95,10 @@ class BackgroundWorker {
     tabRecords.enabled = false;
     this.data[this.currentTab] = tabRecords;
   }
+  clearData(){
+    const requestsLength = this.data[this.currentTab].requests.length;
+    this.data[this.currentTab].requests.splice(0, requestsLength)
```
revathskumar: why don't we just do `this.data[this.currentTab] = {}` ?

app/background.ts
```diff
@@ -59,14 +63,19 @@ class BackgroundWorker {
 
   callback(details:any) {
     const tabRecords = this.data[this.currentTab];
-    if (tabRecords.enabled && this.currentTab === details.tabId) {
-      tabRecords.requests.push(details);
-      this.data[this.currentTab] = tabRecords;
-    }
+    tabRecords.requests.slice(0)  //possible fix for displaying proper badge text
```
revathskumar: why we need to do this? `.slice` won't update the original array.

amit-mb: Yea, realised that and used ```tabRecords.requests = []``` . Will commit that

app/actions/index.ts
```diff
@@ -0,0 +1,31 @@
...
+export const  ERROR = "ERROR";
+export const  DEFAULT_ACTION = "DEFAULT";
+export const CLEAR_REQUESTS = "CLEAR_REQUESTS"
+export const RELOAD_REQUESTS = "RELOAD_REQUESTS"
+export const  UPDATE_FIELD = "UPDATE_FIELD";
+export const  UPDATE_FIELDS = "UPDATE_FIELDS";
+
+// Action Creators
+
+export function startListening (enabledStatus:PopUpState["enabled"]){
```
kgrz: If you find yourself trying to access a type's member to use that member's type, it's a good idea to alias that type and use that instead. This could be a type of `PopUpStatus`, and use that both in `PopUpState` here.

app/types/index.ts
```diff
@@ -0,0 +1,23 @@
+import {RequestObj} from './../request_list'
+export type POPUP_PROPS = {
```
kgrz: Why a type instead of an interface here?

app/types/index.ts
```diff
@@ -0,0 +1,23 @@
...
+}
+
+export interface PopUpState {
+  enabled: boolean;
+  errorMessage ?: string;
+  requests: object;
+  tabUrl ?: string
+}
+
+export interface Action extends PopUpState{
```
kgrz: Why does action need to extend popupstate?

app/background.ts
revathskumar: ya `this.data[this.currentTab].requests = [];` good enough

app/types/index.ts
amit-mb: since Action includes all the properties of popupstate and an additional `type` property, extended it.. Any suggestions?

app/types/index.ts
amit-mb: Right. Fixed that. Thanks!

app/types/index.ts
kgrz: No, was just wondering. In general, if this is a redux action, it need not have to be a global match all type.

PR #33, opened 2018-02-07
Remove existing css and update it with new ui design css

cgirish:


cgirish:
Screenshots for reference -
extension
blank-extension
with-data

PR #35, opened 2018-02-13
Filter request urls and method

amit-mb:
The css for react-table isn't included in this PR. Will  provide the related CSS to @cgirish so that he can commit to https://github.com/code-mancers/interceptor/pull/33

Update : Added react-table CSS for better viewablilty


app/content.ts
```diff
@@ -0,0 +1,75 @@
+import * as $ from 'jquery'
```
kgrz: Why jquery?

app/__tests__/popup.test.tsx
```diff
@@ -0,0 +1,103 @@
+///<reference path='./../../node_modules/@types/enzyme/index.d.ts'/>
+///<reference path='./../../node_modules/@types/jasmine/index.d.ts'/>
+///<reference path='./../../node_modules/@types/enzyme-adapter-react-15/index.d.ts'/>
+///<reference path='./../../node_modules/@types/jest/index.d.ts'/>
+import * as React from 'react';
+import { shallow } from 'enzyme';
+import {Popup} from './../popup';
+
+import * as MessageService from '../message_service';
```
kgrz: I’d suggest writing tests in JS. Useless boilerplate of TS gets in the way of understanding the test itself, IMO

app/content.ts
```diff
@@ -0,0 +1,75 @@
...
+  method:string;
+  requestId:number,
+  timeStamp:number
+}
+class Intercept {
+  requestDetail: object;
+  constructor() {
+    this.requestDetail = {};
+    this.startMessageListener = this.startMessageListener.bind(this);
+    this.injectScripts = this.injectScripts.bind(this);
```
kgrz: You don’t need to bind if you’re using arrow functions. Check the difference.

app/content.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+  startMessageListener = () => {
+    chrome.runtime.onMessage.addListener((request, _, __) => {
+      if (request.message == "INTERCEPT_REQUEST") {
+        this.initScript(request);
+      }
+    });
+  };
+  injectScripts = () => {
+    var jquery = document.createElement("script");
```
kgrz: Avoid mixing var and rely on const and let. var has hoisting properties very different from the other two.

app/content.ts
kgrz: http://youmightnotneedjquery.com/

app/manifest.json
```diff
@@ -7,7 +7,15 @@
...
   },
-  "browser_action": {
+  "content_scripts": [
+    {
+      "matches": ["http://*/*", "https://*/*"],
+      "js": ["content.js"],
+      "run_at" : "document_idle"
+    }
+  ],
+  "web_accessible_resources": ["lib/sinon.js", "lib/jquery.js"],
```
kgrz: Can you tell the reason for doing this?

app/message_service.ts
```diff
@@ -31,5 +31,8 @@ export function getCount(tabId:number, callback: GenericCallback){
 export function clearData(tabId : number){
   chrome.runtime.sendMessage({message : "CLEAR_DATA", tabId})
 }
+export function interceptRequests(tabId:number,url:string, method:string, statusCode:number){
```
kgrz: It's a good practice to use an object for options if there are > 2 arguments. It can get confusing when removing the values. It's okay in TS, but goes south very quickly in JS.

app/popup.tsx
```diff
@@ -35,17 +33,21 @@ export class Popup extends React.Component<POPUP_PROPS & DispatchProps, {}  >{
     return !tabUrl || isChromeUrl(tabUrl);
   };
 
-  handleClick = (_: React.MouseEvent<{}>) : void => { 
+  interceptRequests= (url:string, method:string, statusCode:number) =>  {
+    MessageService.interceptRequests(this.props.tabId, url,method,statusCode)
+  }
+
+  handleClick = (_: React.MouseEvent<{}>) : void => {
```
kgrz: specify the exact type on which the event is added in `React.MouseEvent<{}>`.

app/request_list.tsx
```diff
@@ -1,30 +1,49 @@
...
-      <td className="method">{request.method}</td> 
-    </tr>
-    )
-  });
-}
-
-export interface RequestObj { requests: Array<chrome.webRequest.WebRequestDetails> }
-
+import ReactTable from 'react-table'
+import * as matchSorter from 'match-sorter'
```
kgrz: Do you need `*` here?

app/request_list.tsx
```diff
@@ -1,30 +1,49 @@
...
-        {renderRequests(props.requests)}
-      </tbody>
-    </table>
-  );
+  const columns = [{
+    Header: 'Request URL',
+    accessor: 'url',
+    filterable : true,
+    filterMethod: (filter, rows) =>{
+      console.log(rows)
```
kgrz: remove this?

app/request_list.tsx
```diff
@@ -1,30 +1,49 @@
...
+      return matchSorter(rows, filter.value, { keys: ["url"] , threshold: matchSorter.rankings.CONTAINS})
+    },
+    filterAll: true
+  }, {
+    Header: 'Method',
+    accessor: 'method',
+    filterable : true,
+    filterMethod: (filter, row) => row[filter.id] === filter.value,
+    Filter: ({ filter, onChange }) => <select
```
kgrz: Move these out. It's pretty hard to understand the object structure. And please use either eslint (with not too strict configs. Avoid airbnb one), or prettier which, if integrated properly with your editor, will automatically format the files in a consistent manner.

app/request_list.tsx
```diff
@@ -1,30 +1,49 @@
...
+        value={filter ? filter.value : 'all'}
+      >
+      <option value="GET">GET</option>
+      <option value="POST">POST</option>
+      <option value="OPTIONS">OPTIONS</option>
+    </select>,
+  }, {
+    id: 'timeStamp', // Required because our accessor is not a string
+    Header: 'Time Stamp',
+    accessor: d => new Date(d.timeStamp).toTimeString() // Custom value accessors!
```
kgrz: What does this do?

tsconfig.json
```diff
@@ -14,10 +14,18 @@
...
+    "skipLibCheck": true,
 		"lib": [ "es2016", "es2015", "dom", "dom.iterable" ]
 	},
+  "file_exclude_patterns" : ["*.js, *.js.map"],
 	"include": [
 		"./app/**/*"
-	]
+  ],
+  "exclude": [
+		"./app/_*"
```
kgrz: remove this. Don't add files with `_` prefixes

app/content.ts
amit-mb: Sure. Was using it to just to remove the existing script tag from DOM. Will use native API to do that

app/manifest.json
amit-mb: I am injecting jquery and sinon to DOM. https://github.com/code-mancers/interceptor/blob/c960288208613e7beffa1b1ada63f371a87401eb/app/content.ts#L25-L33 . Content-scripts are loaded in a context different from DOM's context, So packed those scripts in extension itself rather than using a CDN. And again, I could avoid jquery for making Ajax requests

kgrz: I understand you want these scripts to run, but won't webpack automatically bundle these together?

app/request_list.tsx
amit-mb: A column to display the arrival time for requests. Not needed I guess?

app/manifest.json
kgrz: Ah, okay. this usage is different.

kgrz: However, there's a potential problem. What if the target URL already includes jquery? possibly of a different version?

amit-mb: Yeah, that's something to consider. Maybe lets avoid jquery completely as it becomes too heavy for our use case and use something like axios and check for conflicts before injecting?

app/request_list.tsx
```diff
@@ -2,11 +2,41 @@
...
 import ReactTable from "react-table";
 import * as matchSorter from "match-sorter";
+import * as $ from "jquery";
 export interface RequestObj {
   requests: Array<chrome.webRequest.WebRequestDetails>;
   handleIntercept: React.MouseEventHandler<HTMLButtonElement>;
 }
 const RequestList = (props: RequestObj) => {
+  const InterceptComponent = props => {
+    let responseText = "{msg:hello}";
```
amit-mb: I have hardcoded the text here on L12. Since this component is a Stateless component, I would not be able to access its state or refs and hence cannot record the `event.target.value` of the textarea to be passed on to props.handleIntercept when the Intercept button is clicked. Should I make it a stateful component to achieve this?

amit-mb:
Last commit 57ad2104dad26b0736e257901564f35e9f974f62 of this PR fixes #38

PR #39, opened 2018-02-23
Add checkboxes, track checks, Fix enabledStatus button

amit-mb:
This PR fixes the Start Listening Button that wasn't persisting each time the popup was loaded. Checkboxes are added to each column. Clicked checks are tracked using an unique id(requestId).


app/modules/recordings.ts
```diff
@@ -1,5 +1,5 @@
 import {PopUpState, Action} from '../types'
-const INITIAL_POPUP_STATE : PopUpState = { enabled: false , errorMessage: "",requests: []}
+const INITIAL_POPUP_STATE : PopUpState = { enabled: false , errorMessage: "", requests: [], selectedReqIds: [], checked : false}
```
revathskumar: As discussed, why this `checked` is needed? please remove

app/modules/recordings.ts
```diff
@@ -9,15 +9,30 @@ export const reducer = (state = INITIAL_POPUP_STATE, action: Action) => {
...
-    case actionType.CLEAR_REQUESTS : return {...state, requests: [] }
+    case actionType.CLEAR_REQUESTS :
+      return {...state, requests: [] }
+    case actionType.TOGGLE_CHECKBOX:
+      let selectedReqIds = state.selectedReqIds;
+      if (action.checked) {
+        selectedReqIds = [ ...selectedReqIds, action.reqId ]
+        return {...state , selectedReqIds, checked : !action.checked }
+      } else {
+        removeReqId(state, action)
```
revathskumar: As discussed don't pass the whole state and action. Pass only the minimum needed items.

app/popup.tsx
```diff
@@ -31,6 +33,9 @@ const isChromeUrl = (url: string) => {
 
 export class Popup extends React.Component<POPUP_PROPS & DispatchProps, {}> {
   componentWillMount() {
+    MessageService.getEnabledStatus(this.props.tabId, (enabled: boolean) => {
```
revathskumar: You are doing this anyway in the `app.tsx`. No need to do this here. 
This will just slow down and make unnecessay messaging

app/app.tsx
```diff
@@ -21,10 +21,9 @@ chrome.tabs.query(queryParams, tabs => {
   if (typeof id === "undefined" || typeof url === "undefined") return;
 
   MessageService.getEnabledStatus(id, (enabled: boolean) => {
```
revathskumar: you are getting the status here. So use it and update the state. there is two ways to do this.
1. do a `store.dispatch()` action needed to update the state.
2. export the initial data and use `Object.assign` or `spread` operator to update the enabled state and pass it as initialState in `store()`.

app/request_list.tsx
```diff
@@ -32,17 +34,58 @@ const RequestList = (props: RequestObj) => {
...
+            return false
+          }
+
+        }
+
+        return (
+          <input
+            type="checkbox"
+            className="checkbox"
+            checked={checkedBool(original.requestId, props.selectedReqIds) || false}
```
revathskumar: why not `props.selectedReqIds.indexOf(original.requestId) > 0` ?

app/request_list.tsx
```diff
@@ -32,17 +34,58 @@ const RequestList = (props: RequestObj) => {
...
     {
-      Header: "Request ID",
-      accessor: "requestId",
-      filterable: true
+      id: "checkbox",
+      accessor: "",
+      Cell: ({original}) => {
+        const presentReqId = (props, original) => {
+          let RequestId = -1;
+          props.requests.map( (request, index:number) => {
```
revathskumar: Just use ` props.requests.find()`

app/types/index.ts
```diff
@@ -4,31 +4,35 @@ export interface POPUP_PROPS {
   enabled : boolean,
   requests : Array<any>,
   errorMessage : string,
-  responseText : string,
-  selectStatusCode :  string
+  selectedReqIds : Array<number>,
+  checked : Array<boolean>
```
revathskumar: remove this checked? Why this is required? I guess we already discussed this.

app/modules/recordings.ts
amit-mb: Tried removing it, we later check the value of checked below as 
https://github.com/code-mancers/interceptor/blob/0c253454b3d57d02aa2e7abfee18d50d2f1a573e/app/modules/recordings.ts#L23

If removed, the ID's wont get added to the array

app/modules/recordings.ts
amit-mb: The newState is being returned from the function on L37 as https://github.com/code-mancers/interceptor/blob/0c253454b3d57d02aa2e7abfee18d50d2f1a573e/app/modules/recordings.ts#L37 and all the operations relating to state are done inside the function unlike earlier as 
```js
{...state, removeReqId(state, action) } 
```

Still recommend to do the earlier way as in 
```js
{...state, removeReqId(state.selectedReqIds, action.reqId) } 
``` 
?

app/popup.tsx
amit-mb: Sure 👍 . Will pass the same from `app.tsx`

app/popup.tsx
```diff
@@ -73,6 +76,15 @@ export class Popup extends React.Component<POPUP_PROPS & DispatchProps, {}> {
     this.props.clearFields();
   };
 
+  handleCheckToggle = (reqId:number, presentCheckedState:boolean) => {
+    this.props.handleCheckToggle(reqId, presentCheckedState)
+  };
+
+  handleCheckedRequests = (requests:Array<any>) =>{
+    const tabId:number = this.props.tabId
```
revathskumar: Why assign to a `const`? pass the `this.props.tabId` directly.

app/content.ts
```diff
@@ -14,32 +14,70 @@ class Intercept {
...
+  };
   injectScripts = () => {
     let sinon = document.createElement("script");
     sinon.defer = false;
     sinon.src = chrome.extension.getURL("./lib/sinon.js");
     (document.head || document.documentElement).appendChild(sinon);
   };
   initScript = (request: requestObject) => {
-    var actualCode = `
+    let actualCode = `
```
revathskumar: 👍

app/popup.tsx
revathskumar: marking this as a non blocker and approving.

PR #42, opened 2018-03-06
Fix to prevent start listening on error pages

amit-mb:


app/popup.tsx
```diff
@@ -58,17 +58,17 @@ export class Popup extends React.Component<POPUP_PROPS & DispatchProps, {}> {
     if (this.props.enabled) {
       MessageService.disableLogging(this.props.tabUrl, this.props.tabId);
       this.props.updateField("enabled", false);
-    } else {
+    }
+    else if (this.isUrlInValid(this.props.tabUrl)) {
```
revathskumar: no need for else. you can move this block up and return;

revathskumar:
@amit-mb We already have a test case. So we don't need to add new one instead we might need to update one. Once you fix the other comment. lets discuss.

app/__tests__/popup.test.jsx
```diff
@@ -113,6 +112,7 @@ describe("Popup", () => {
       expect(localProps.errorNotify).toHaveBeenCalledWith(
         "Cannot Start Listening on chrome://version"
       );
+      expect(localProps.enabled).toEqual(false)
```
revathskumar: instead of this **a better test** will be to make sure we are not calling `disableLogging` or `getRequests`

app/__tests__/popup.test.jsx
```diff
@@ -113,6 +112,10 @@ describe("Popup", () => {
       expect(localProps.errorNotify).toHaveBeenCalledWith(
         "Cannot Start Listening on chrome://version"
       );
+      expect(localProps.enabled).toEqual(false);
```
revathskumar: this is not required.

app/__tests__/popup.test.jsx
```diff
@@ -113,6 +112,10 @@ describe("Popup", () => {
...
+      expect(MessageService.enableLogging).toHaveBeenCalledTimes(0)
+      expect(MessageService.disableLogging).toHaveBeenCalledTimes(0)
```
revathskumar: `expect(MessageService.disableLogging).not.toHaveBeenCalled()`

app/__tests__/popup.test.jsx
```diff
@@ -113,6 +112,10 @@ describe("Popup", () => {
       expect(localProps.errorNotify).toHaveBeenCalledWith(
         "Cannot Start Listening on chrome://version"
       );
+      expect(localProps.enabled).toEqual(false);
+      expect(MessageService.enableLogging).toHaveBeenCalledTimes(0)
```
revathskumar: Instead check for `getRequests`.

app/__tests__/popup.test.jsx
```diff
@@ -113,6 +112,10 @@ describe("Popup", () => {
...
+      expect(MessageService.disableLogging).toHaveBeenCalledTimes(0)
+      expect(localProps.startListening).toHaveBeenCalledTimes(0);
```
revathskumar: Not required.

PR #43, opened 2018-03-08
Attach-event-handlers-to-set-props

amit-mb:


app/Intercept_Components/InterceptTextBox.tsx
```diff
@@ -1,25 +1,51 @@
 import * as React from 'react'
 
 export const InterceptTextBox = props => {
-  console.log(props)
   let responseText = "{msg:hello}";
+  const checkResponseExists = (index) => {
+    if(props.ResponseText[index]){
```
revathskumar: Why are we still using `index`? we already had discussion and decided to go with request id.

app/Intercept_Components/InterceptTextBox.tsx
```diff
@@ -1,25 +1,51 @@
...
+      return props.ResponseText[index]
+    }else{
+      return responseText
```
revathskumar: name it as `defaultResponseText`.

app/Intercept_Components/InterceptTextBox.tsx
```diff
@@ -1,25 +1,51 @@
...
+    }
+  }
+  const checkStatusCode = (index) => {
```
revathskumar: Please don't define methods like this. If you need instance method use the `class` syntax.

app/Intercept_Components/InterceptTextBox.tsx
```diff
@@ -1,25 +1,51 @@
...
         value="Intercept"
         onClick={props.handleIntercept.bind(
           this,
           props.rowProps.url,
           props.rowProps.method,
-          responseText,
-          200
-        )}
+          checkResponseExists(props.rowProps.checkbox.requestId),
+          checkStatusCode(props.rowProps.checkbox.requestId)
```
revathskumar: instead of calling this method `props.interceptStatus[index] || 200` will do.

app/actions/index.ts
```diff
@@ -34,4 +36,10 @@ export function handleCheckToggle(reqId:number, checked:boolean){
 }
 export function handleCheckedRequests(tabId:number,requests:Array<any>){
   return {type: INTERCEPT_CHECKED , payload : requests}
+}
+export function handleStatusCodeChange(value:string, rowIndex:number){
+  return {type: STATUSCODE_CHANGE , payload : {value, rowIndex} }
```
revathskumar: why are we using rowIndex?

app/content.ts
```diff
@@ -64,21 +64,32 @@ class Intercept {
...
+      // If the filter returns true, the request will not be faked - leave original
+      this.server.xhr.addFilter(function(method, url, async, username, password) {
+        const result = (request.url === url)
+        return !result
+      });
+      this.server.respondWith((xhr, id) => {
+        xhr.respond(200, { "Content-Type": "application/json" },'[${JSON.stringify(request.responseText)}]')
+      })
+    }
+    window.interceptor2 = new sinonSingleHandler(${JSON.stringify(request)});
```
revathskumar: why name as interceptor2?

app/content.ts
```diff
@@ -64,21 +64,32 @@ class Intercept {
...
+      this.server = sinon.fakeServer.create({ logger: console.log });
+      this.server.autoRespond = true;
+      this.server.xhr.useFilters = true;
+      // If the filter returns true, the request will not be faked - leave original
+      this.server.xhr.addFilter(function(method, url, async, username, password) {
+        const result = (request.url === url)
+        return !result
+      });
+      this.server.respondWith((xhr, id) => {
+        xhr.respond(200, { "Content-Type": "application/json" },'[${JSON.stringify(request.responseText)}]')
```
revathskumar: Are we injecting this script for each recorded request.
If that is the case we are doing it wrong. 
once script will be injected for each time we say intercept. all the request selected will be handled by that.

app/popup.tsx
```diff
@@ -80,9 +84,16 @@ export class Popup extends React.Component<POPUP_PROPS & DispatchProps, {}> {
     this.props.handleCheckToggle(reqId, presentCheckedState)
   };
 
+  handleRespTextChange = (value, rowIndex) =>{
+    this.props.handleRespTextChange(value, rowIndex)
```
revathskumar: Don't use rowIndex

app/request_list.tsx
```diff
@@ -81,8 +85,10 @@ const RequestList = (props: RequestObj) => {
       showPaginationTop={false}
       showPaginationBottom={true}
       pageSize={10}
+      freezeWhenExpanded={true}
       SubComponent={row => (
-        <InterceptForm rowProps={row} handleIntercept={props.handleIntercept} />
+        <InterceptForm freezeWhenExpanded={true} rowProps={row} handleIntercept={props.handleIntercept} handleStatusCodeChange={props.handleStatusCodeChange} handleRespTextChange={props.handleRespTextChange} ResponseText={props.ResponseText}
```
revathskumar: use defaultProp and instead of `freezeWhenExpanded={true}` just use `freezeWhenExpanded`

app/Intercept_Components/InterceptTextBox.tsx
amit-mb: I have used requestId as you said and that is what is passed in the call. Will change the names and push the changes

revathskumar:
@amit-mb Adding tests take it as different task.

Issue #34, opened 2018-02-12
Same requests are listed twice sometimes


amit-mb:
~~Happens frequently when <select> filter is used to filter request method~~

revathskumar:
@amit-mb Did you fixed this?

amit-mb:
Working on this currently

PR #46, opened 2018-03-13
Fix logging of multiple requests with same requestId

amit-mb:
This PR fixes https://github.com/code-mancers/interceptor/issues/34

### DIFF
--- a/app/background.ts
+++ b/app/background.ts
@@ -64,10 +64,6 @@ class BackgroundWorker {
 
   callback = (details: any) => {
     const tabRecords = this.data[this.currentTab];
-    chrome.browserAction.setBadgeText({
-      text: `${tabRecords.requests.length}`,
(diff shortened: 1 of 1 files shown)


app/background.ts
```diff
@@ -72,12 +72,15 @@ class BackgroundWorker {
...
-      chrome.browserAction.setBadgeText({
-        text: `${tabRecords.requests.length}`,
-        tabId: details.tabId
-      });
+        if (tabRecords.requests.filter((req) => req.requestId === details.requestId).length > 0) {
+          return
+        }
+        tabRecords.requests.push(details);
+        this.data[this.currentTab] = tabRecords;
+        chrome.browserAction.setBadgeText({
```
revathskumar: Why we need to update the badge twice? beginning of this callback also we are setting  badge.

PR #47, opened 2018-03-14
Persist redux store in a background page

amit-mb:
Use `react-chrome-redux (https://github.com/tshaddix/react-chrome-redux)` package to create redux store in a background page. All fields such as text boxes and typed responses are persisted even when the popup is closed.


app/popup.tsx
```diff
@@ -58,10 +64,10 @@ export class Popup extends React.Component<POPUP_PROPS & DispatchProps, {}> {
     if (this.props.enabled) {
       MessageService.disableLogging(this.props.tabUrl, this.props.tabId);
       this.props.updateField("enabled", false);
-    }
-    else {
+    } else {
       MessageService.getRequests(this.props.tabId, requests => {
         MessageService.enableLogging(this.props.tabUrl, this.props.tabId);
+        console.log("GOT REQUESTS", requests);
```
revathskumar: remove log

app/popup.tsx
```diff
@@ -44,6 +46,10 @@ export class Popup extends React.Component<POPUP_PROPS & DispatchProps, {}> {
     MessageService.getRequests(this.props.tabId, requests => {
       this.props.updateField("requests", requests);
     });
+    if (!this.isUrlInValid(this.props.tabUrl)) {
+      this.props.updateField("errorMessage", "");
```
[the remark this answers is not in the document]
amit-mb: Tried doing that. `this.props.updateField("errorMessage", "")` doesn't get called on `ComponentWillUnmount`. It may be that the component gets unmounted before the call gets executed. Hence called it on `componentWillMount`

app/popup.tsx
```diff
@@ -44,6 +44,15 @@ export class Popup extends React.Component<POPUP_PROPS & DispatchProps, {}> {
     MessageService.getRequests(this.props.tabId, requests => {
       this.props.updateField("requests", requests);
     });
+
+    MessageService.getEnabledStatus(this.props.tabId, (enabledStatus:boolean) =>{
+      this.props.updateField("enabled", enabledStatus)
+    })
+
+    if (!(this.isUrlInValid(this.props.tabUrl))) {
+      this.props.updateField("errorMessage", "");
```
revathskumar: why isn't this moved to unmount?

revathskumar: okey didn't saw the reply wait

app/popup.tsx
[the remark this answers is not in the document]
revathskumar: its componentWillUnmount. 
Any error its showing?

app/popup.tsx
```diff
@@ -44,6 +44,15 @@ export class Popup extends React.Component<POPUP_PROPS & DispatchProps, {}> {
     MessageService.getRequests(this.props.tabId, requests => {
       this.props.updateField("requests", requests);
     });
+
+    MessageService.getEnabledStatus(this.props.tabId, (enabledStatus:boolean) =>{
+      this.props.updateField("enabled", enabledStatus)
+    })
+
+    if (!(this.isUrlInValid(this.props.tabUrl))) {
```
revathskumar: if this is in mount we don't need any checking here.

app/popup.tsx
```diff
@@ -44,6 +44,15 @@ export class Popup extends React.Component<POPUP_PROPS & DispatchProps, {}> {
...
       this.props.updateField("requests", requests);
     });
+
+    MessageService.getEnabledStatus(this.props.tabId, (enabledStatus:boolean) =>{
+      this.props.updateField("enabled", enabledStatus)
+    })
+
+    if (!(this.isUrlInValid(this.props.tabUrl))) {
+      this.props.updateField("errorMessage", "");
+      return;
```
revathskumar: why this return?

amit-mb: Suppose a user tries to Start Listening to requests on chrome:// pages, it gives us an error by checking whether `this.props.tabUrl` is a chrome URL. Now, since the redux state gets persisted across all tabs, this check is needed on `componentWillMount` for all tabs to remove the earlier error generated from this [line](https://github.com/code-mancers/interceptor/pull/47/files/48157dc30868f91f1b6f5bf2e64eb97c3fcc4f8c#diff-9744d706869f0ab0f5de1332df64040cR116)

amit-mb: Yes, the `return` isn't required. Removed that 👍

revathskumar: What ever the error is, we can clear the errorMessage when the user reopen the popup.
Its is not just meant for validURl. Hence we don't need a check there.

PR #49, opened 2018-03-16
Add ability to intercept on page-refresh

amit-mb:
This PR adds ability to intercept the requests for a page refresh without having to click the "Intercept" button manually. It also addresses the issue of interception across many tabs (If a request is intercepted on one tab, it would respond with a fake response even when the same request is sent through a different tab).


app/redux_bg.ts
```diff
@@ -8,3 +8,9 @@ const createStore = store({ ...INITIAL_POPUP_STATE });
 wrapStore(createStore, {
   portName: 'INTERCEPTOR',
 })
+
+chrome.tabs.onUpdated.addListener((tabId, changeInfo, tab) => {
```
revathskumar: why this is in `redux` file. suppose to be in other background page.

amit-mb: I have defined two background pages. One for listening and handling our `webRequest`'s and the other for initiating the redux store in background. The `manifest.json` file is as below

https://github.com/code-mancers/interceptor/blob/2b92a5ae1f67b2cf3c2097558640a7676dc76909/app/manifest.json#L8

revathskumar: I understand that. I am asking why this `chrome.tabs.onUpdated.addListener` done in redux background page. It should happen in other background page where all eventlisteners are handled.

PR #52, opened 2018-03-16
Return early if presenttabId doesn't match request tabId

amit-mb:
This PR returns early from the content-script and hence prevent creating unused variables if none of the request's tabId match the present tabId.

### DIFF
--- a/app/content.ts
+++ b/app/content.ts
@@ -36,7 +36,7 @@ class Intercept {
     });
   };
   interceptSelected = (selectedReqs: Array<requestObject>) => {
-    if (selectedReqs.requestsToIntercept.length < 1) {
+    if (selectedReqs.requestsToIntercept.length < 1 || !selectedReqs.tabId || selectedReqs.requestsToIntercept.find( (req) => req.tabId !== selectedReqs.tabId )){
       return;
     }
     let responseTexts = selectedReqs.responseText || {};
--- a/app/content.ts
+++ b/app/content.ts
@@ -15,25 +15,27 @@ class Intercept {
       portName: "INTERCEPTOR"
     });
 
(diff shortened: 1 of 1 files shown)


app/content.ts
```diff
@@ -36,7 +36,7 @@ class Intercept {
     });
   };
   interceptSelected = (selectedReqs: Array<requestObject>) => {
-    if (selectedReqs.requestsToIntercept.length < 1) {
+    if (selectedReqs.requestsToIntercept.length < 1 || !selectedReqs.tabId || selectedReqs.requestsToIntercept.find( (req) => req.tabId !== selectedReqs.tabId )){
```
revathskumar: Isn't `selectedReqs.requestsToIntercept.find( (req) => req.tabId !== selectedReqs.tabId )` will return `true` when selectedReqs.requestsToIntercept is not empty? 

Do we really need this condition?

amit-mb: If the above check is not present, Consider a case where a request is intercepted is intercepted in one tab. On opening the other tab, the variables declared below will unnecessarily be created. 

https://github.com/code-mancers/interceptor/blob/7ca039b5954729b7a78bba68ea95d59db3c86689/app/content.ts#L42-L62

These requests will anyway not be intercepted during the check we do in `AddFilter` callback as below.
https://github.com/code-mancers/interceptor/blob/7ca039b5954729b7a78bba68ea95d59db3c86689/app/content.ts#L81-L86
 By checking the condition early, we are preventing the creation of unused variables.

amit-mb: You can `console.log(selectedReqs)` here and check the program flow for different tabs. For a different tab, the `log` isn't printed which means it returns early for requests of other tab

PR #58, opened 2018-03-20
Store current page number and page size of react-table in redux store

### LINKED ISSUE #55: Store the pagination values in redux store (1)
Store the pagination values like page number and no of rows per page in redux store.

amit-mb:
This PR stores the current  page number and page-size of react-table with tabId as the key in redux store so that the page numbers and size do not reset to default values when we change tabs or close the popup.

Fixes #55

### DIFF
--- a/app/modules/recordings.ts
+++ b/app/modules/recordings.ts
@@ -1,34 +1,76 @@
-import {POPUP_PROPS, Action} from '../types'
-export const INITIAL_POPUP_STATE : POPUP_PROPS = { enabled: false , errorMessage: "", requests: [], checkedReqs : {}, responseText: {}, statusCodes: {}, contentType: {}}
+import {POPUP_PROPS, Action} from "../types";
+export const INITIAL_POPUP_STATE: POPUP_PROPS = {
+  enabled: false,
+  tabId: -1,
+  tabUrl: "",
+  errorMessage: "",
+  requests: [],
+  checkedReqs: {},
+  responseText: {},
+  statusCodes: {},
+  contentType: {},
+  PageDetails: {}
+};
 
 //ACTION CONSTANTS
-import * as actionType from "../actions"
+import * as actionType from "../actions";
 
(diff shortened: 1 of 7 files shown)


app/modules/recordings.ts
```diff
@@ -1,34 +1,76 @@
...
+        ...state,
+        PageDetails: {
+          ...state.PageDetails,
+          [action.payload.tabId]: {
+            ...state.PageDetails[action.payload.tabId],
+            currentPageNumber: action.payload.currentPageNumber
+          }
+        }
+      };
+    case actionType.ROW_SIZE_CHANGE:
```
revathskumar: As discussed, These two can be clubbed into single action.

app/popup.tsx
```diff
@@ -127,6 +141,10 @@ export class Popup extends React.Component<POPUP_PROPS & DispatchProps, {}> {
           statusCodes={this.props.statusCodes}
           handleContentTypeChange={this.props.handleContentTypeChange}
           contentType={this.props.contentType}
+          handlePageNumberChange = {this.props.handlePageNumberChange}
+          handleRowSizeChange = {this.props.handleRowSizeChange}
+          PageDetails = {this.props.PageDetails}
```
revathskumar: Need to send only the `pageDetails` of the currentTab.

app/request_list.tsx
```diff
@@ -85,11 +94,23 @@ const RequestList = (props: RequestObj) => {
       showPagination={true}
       showPaginationTop={false}
       showPaginationBottom={true}
-      pageSize={10}
-      freezeWhenExpanded
+      defaultPageSize={10}
+      page={props.PageDetails[props.tabId] ? props.PageDetails[props.tabId].currentPageNumber : 0}
```
revathskumar: should the default pageNumber 0 or 1?

amit-mb: The page numbers start with `0`. It uses `index`. However, the pageNumber is displayed as `1` on view.

Issue #56, opened 2018-03-19
Intercept button and messages (2)


revathskumar:
* Move intercept button to top left outside table.
* Once the intercept script is injected show some INTENT to user via success message, on error show error message.
* Disable the intercept button when user didn't selected any requests.

amit-mb:
The above mentioned 3 points are done and #62 is merged with master

PR #59, opened 2018-03-20
Move Intercept All button outside the table

amit-mb:
This PR address first and third item of #56

### DIFF
--- a/app/InterceptAllButton.tsx
+++ b/app/InterceptAllButton.tsx
@@ -1,16 +1,12 @@
 import * as React from "react";
 
 export const InterceptAllButton = props => {
-  const enabledRequests = props.requests
-  .filter(request => {
-    return props.checkedReqs[request.requestId];
-  })
   return (
     <button
     id="intercept-all-btn"
-    disabled={!enabledRequests.length}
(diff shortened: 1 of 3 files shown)


app/InterceptAllButton.tsx
```diff
@@ -0,0 +1,24 @@
+import * as React from "react";
+
+export const InterceptAllButton = props => {
```
revathskumar: This component doesn't need to know about `checkedReqs` or `enabledRequests`.
It should accept a eventHandler and `disabled` props

app/request_list.tsx
```diff
@@ -67,27 +68,20 @@ const RequestList = (props: RequestObj) => {
...
+
+  const enabledRequests = props.requests.filter(request => {
+      return props.checkedReqs[request.requestId];
+    })
+
   return (
+    <div>
+    <InterceptAllButton
+    disabled={!enabledRequests.length}
+    handleCheckedRequests={props.handleCheckedRequests(enabledRequests)}
```
revathskumar: this gets called when it renders, not on `onClick`. please test this

Issue #60, opened 2018-03-21
Interception script runs before the sinon script is injected 


amit-mb:
On certain websites (happened with me for `https://github.com`), the interception script gets injected and runs even before loading of sinon.js script and hence throws the following error.

For some reason, the error is not thrown when testing on [jsbin](https://output.jsbin.com/bulunet)

https://github.com/code-mancers/interceptor/blob/f850ed4c0f4930c891190a12c9e3c0c779068c5e/app/content.ts#L64-L99

Specifically, the above script gets executed before parsing `sinon.js`

The following error s thrown:

![screen shot 2018-03-21 at 11 04 30 am](https://user-images.githubusercontent.com/5921327/37695857-6eb78798-2cf8-11e8-82ec-7aea85c00c36.png)

amit-mb:
Closing this as this is fixed with #62

PR #62, opened 2018-03-21
Display Intercept Success message to view

amit-mb:
Once the intercept script is injected, A simple `SUCCESS` message is displayed on the UI through which the user can verify that the requests are actually intercepted. Also addresses #60

### DIFF
--- a/app/actions/index.ts
+++ b/app/actions/index.ts
@@ -52,6 +52,6 @@ export function handleContentTypeChange(value:string, requestId:number){
 export function handlePaginationChange(value: string, tabId:number, field:string) {
   return {type: PAGINATION_CHANGE, payload: {field, value, tabId}};
 }
-export function sendSuccessMessage(tabId:number){
-  return {type : INTERCEPT_SUCCESS , payload : { tabId } }
+export function sendSuccessMessage(){
+  return {type : INTERCEPT_SUCCESS }
 }
--- a/app/actions/index.ts
+++ b/app/actions/index.ts
@@ -13,7 +13,7 @@ export const STATUSCODE_CHANGE = "STATUSCODE_CHANGE"
 export const RESP_TEXT_CHANGE = "RESP_TEXT_CHANGE"
 export const CONTENT_TYPE_CHANGE = "CONTENT_TYPE_CHANGE"
 export const PAGINATION_CHANGE = "PAGINATION_CHANGE"
-export const INTERCEPT_SUCCESS = "INTERCEPT_SUCCESS"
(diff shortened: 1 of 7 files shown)


app/content.ts
```diff
@@ -103,16 +101,20 @@ class Intercept {
     script.type = "text/javascript";
     script.textContent = selectedInterceptCode;
     (document.head || document.documentElement).appendChild(script);
+    }
+    this.injectScripts(runInterceptor);
+    bg_store.dispatch({type : "INTERCEPT_SUCCESS", payload : { tabId : selectedReqs.tabId } })
```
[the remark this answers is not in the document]
revathskumar: Also add an action for update message

app/actions/index.ts
```diff
@@ -50,4 +51,7 @@ export function handleContentTypeChange(value:string, requestId:number){
 }
 export function handlePaginationChange(value: string, tabId:number, field:string) {
   return {type: PAGINATION_CHANGE, payload: {field, value, tabId}};
+}
+export function sendSuccessMessage(tabId:number){
```
revathskumar: This should accept message as well

revathskumar: As discussed there is no need for tabId. this should not depend on tabId. anyway we will be clear this on mounting.

app/content.ts
```diff
@@ -38,28 +48,26 @@ class Intercept {
...
     }
-    this.injectScripts();
+    this.injectScripts(this.runInterceptor(selectedReqs));
+    this.store.dispatch(sendSuccessMessage(selectedReqs.tabId))
+  };
+
+  setDefaultValues = (responseField, requestsToIntercept, defaultResponseValue) => {
+    if (Object.keys(responseField).length === 0 && responseField.constructor === Object) {
+      return responseField = requestsToIntercept.map(req => {
+        responseField[req.requestId] = defaultResponseValue;
```
revathskumar: This is a wrong usage of `map`. What you are trying to return from this map?

app/popup.tsx
```diff
@@ -110,6 +110,13 @@ export class Popup extends React.Component<POPUP_PROPS & DispatchProps, {}> {
     this.props.handlePaginationChange(newPageNo_rowSize, tabId, field)
   }
 
+  displayInterceptMessage = () => {
+   return Object.keys(this.props.interceptData).map( tabId =>
+     this.props.tabId === Number(tabId) && this.props.interceptData[this.props.tabId] === "INTERCEPT_SUCCESS" ?
+    <a id="success-msg">{this.props.interceptData[this.props.tabId]}</a> :
```
revathskumar: Why this is a `<a>`.
No need for all these complications.
`message && <div>{message}</div>`

Issue #66, opened 2018-03-26
Bug : sinon is not defined


amit-mb:
We get a error is console saying `sinon is not defined` (happens both on chrome and firefox) which hints that the sinon script isn't injected and parsed properly before the interception process. Need to debug this

amit-mb:
Closing as the issue got addressed in #70

PR #70, opened 2018-03-26
Fix sinon-script not being loaded before interception

amit-mb:
#66. Fix : Executed the call back through a anonymous function unlike earlier where the function call was fired before the `sinon-script` was loaded


app/content.ts
```diff
@@ -94,7 +94,7 @@ class Intercept {
            });
            this.server.respondWith((xhr, id) => {
              const respondUrl = requestArray.requestsToIntercept.find((request) => {
-              if(xhr.url === request.url){
+              if(xhr.url === request.url && requestArray.responseText[request.requestId] && requestArray.contentType[request.requestId] && requestArray.statusCodes[request.requestId]){
```
revathskumar: why we need this big condition? we have default values for all these right?

PR #72, opened 2018-03-26
Fix cursor position moving to the end on textarea onChange

### LINKED ISSUE #68: The cursor in the text area shouldn't move to the end while typing
The cursor in in response text field keeps moving to the end whenever a character is pressed(may be the component re-renders each time `onChange` event is fired).

amit-mb:
Attempt to fix #68


revathskumar:
@amit-mb please resolve the conflicts.

PR #77, opened 2018-03-27
Show the entire URL as title

amit-mb:
Show the entire URL as title. Useful when the user wants to know the entire URL on hovering instead of dragging the table across

### DIFF
--- a/app/components/request_list.tsx
+++ b/app/components/request_list.tsx
@@ -23,8 +23,7 @@ const RequestList = (props: RequestObj) => {
   const columns = [
     {
       Header: "Request URL",
-      className: "url",
-      accessor: "url",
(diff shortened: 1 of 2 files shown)


app/components/request_list.tsx
```diff
@@ -22,7 +22,7 @@ const RequestList = (props: RequestObj) => {
   const columns = [
     {
       Header: "Request URL",
-      accessor: "url",
+      Cell: ({original}) => <div style={{flex: '100 0 auto', width: '100px'}}className="url" title={original.url}>{original.url}</div>,
```
revathskumar: possible to move the style to class `url`?

amit-mb: Yeah

PR #78, opened 2018-03-27
Fix#76  requests not clearing when response form subcomponent is mounted

### LINKED ISSUE #76: Requests do not get cleared when the response form is open
On clicking the clear button when the response form is open, the requests don't get clear immediately. Instead they are cleared after closing the form or when `componentMounts`

requestsnotclearing

amit-mb:
Fixes #76


app/components/request_list.tsx
```diff
@@ -108,10 +108,9 @@ const RequestList = (props: RequestObj) => {
       pageSize={props.PageDetails[props.tabId] ? props.PageDetails[props.tabId].currentRowSize : 10}
       onPageChange={changedPageNo => props.handlePaginationChange(changedPageNo, props.tabId, "currentPageNumber")}
       onPageSizeChange={changedRowSize => props.handlePaginationChange(changedRowSize, props.tabId, "currentRowSize")}
-      freezeWhenExpanded={true}
+      collapseOnDataChange={false}
```
revathskumar: why we need to pass this `collapseOnDataChange`? since its false, we can skip right?

amit-mb: No. The default value is set to true. check [this](https://github.com/react-tools/react-table#props)

PR #79, opened 2018-03-28
Move fonts to local

amit-mb:


revathskumar:
I think @cgirish is the better person to review this PR.

PR #81, opened 2018-03-28
Respond to empty fake responses

amit-mb:
Earlier if the response text specified in the `textarea` was blank, such requests were responded with default text. This PR allows interceptor to mock empty `fake` responses


app/content/content.ts
```diff
@@ -56,7 +56,7 @@ class Intercept {
 
   setDefaultValues = (responseField, requestsToIntercept, defaultResponseValue) => {
       requestsToIntercept.forEach(req => {
-        if (!(responseField[req.requestId] && responseField[req.requestId].trim())) {
```
revathskumar: This change is not enough.
* we need to remove https://github.com/code-mancers/interceptor/pull/81/files#diff-438094ec72049283bd76620ef44fe1d3R70
* Then in onChange event of textarea,

amit-mb:
Closing this as this branch was not based on `master`.  Will open another PR

Issue #80, opened 2018-03-28
Add ability to disable interceptor


amit-mb:
Once the `Disable` button is clicked, the extension shouldn't mock fake responses and should return the actual response returned by the server for any further requests made

amit-mb:
Closing this as #101 merged into `master`

PR #83, opened 2018-04-02
Add ability to disable interceptor

amit-mb:
This PR addresses #80 . On clicking the 'DISABLE' button, interception is disabled and response from the backend is loaded instead of `sinon fakeServer`


revathskumar:
Closing this since already merged via #101

PR #87, opened 2018-04-03
Add check all btn

amit-mb:
Adds a checkbox as column header which allows the user to check/uncheck all the requests for interception at once

### DIFF
--- a/app/components/request_list.tsx
+++ b/app/components/request_list.tsx
@@ -55,7 +55,19 @@ const RequestList = (props: RequestObj) => {
     },
     {
       id: "checkbox",
-      accessor: "",
+      accessor: "data",
+      Header : (data) => {
+          return <input type="checkbox" className="checkAll" onChange={e => {
+            data.data.map( (val, index) => {
+              if(e.target.checked){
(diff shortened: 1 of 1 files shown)


app/components/request_list.tsx
```diff
@@ -55,9 +55,16 @@ const RequestList = (props: RequestObj) => {
     },
     {
       id: "checkbox",
-      accessor: "",
+      Header : (data) =>
+      <input type="checkbox" className="checkAll" onChange={e =>
+          data.data.forEach( (val) => e.target.checked ?
+          props.handleCheckToggle(val._original.requestId, true) :
```
revathskumar: Why there is method call in `forEach`?

app/components/request_list.tsx
```diff
@@ -55,9 +55,16 @@ const RequestList = (props: RequestObj) => {
     },
     {
       id: "checkbox",
-      accessor: "",
+      Header : (data) =>
+      <input type="checkbox" className="checkAll" onChange={e =>
```
revathskumar: how to determine if the checkbox is checked or not?

PR #91, opened 2018-04-05
Display full URL in subcomponent

### LINKED ISSUE #86: Layout gets screwed on dragging the URL column
On dragging the URL column, the whole table gets moved to right side 

screen shot 2018-04-02 at 4 39 26 pm

amit-mb:
A better approach to fix #86 . Displays the complete URL in expanded state instead of having to look at whole URL by resizing the table
screen shot 2018-04-05 at 2 52 26 pm

### DIFF
--- a/app/components/request_list.tsx
+++ b/app/components/request_list.tsx
@@ -101,6 +101,7 @@ const RequestList = (props: RequestObj) => {
     <ReactTable
       data={props.requests}
       columns={columns}
+      resizable={false}
       showPagination={true}
       showPaginationTop={false}
       showPaginationBottom={true}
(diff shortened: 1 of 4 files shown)


app/components/request_list.tsx
```diff
@@ -101,6 +101,7 @@ const RequestList = (props: RequestObj) => {
     <ReactTable
       data={props.requests}
       columns={columns}
+      resizable={false}
```
revathskumar: I think there is nothing wrong in keep it `resizable`.

cgirish: +1 for @amit-mb, Displays the complete URL in expanded state is better than resizing the table

PR #101, opened 2018-04-09
Add Disable interceptor feature

amit-mb:
This PR addresses #80 . On clicking the 'DISABLE' button, interception is disabled and response from the backend is loaded instead of `sinon fakeServer`

Toggle Switch Screenshot :

screen shot 2018-04-11 at 10 04 23 pm

Off state screenshot  :
screen shot 2018-04-11 at 10 07 52 pm


app/actions/index.ts
```diff
@@ -13,7 +13,8 @@ export const STATUSCODE_CHANGE = "STATUSCODE_CHANGE"
 export const RESP_TEXT_CHANGE = "RESP_TEXT_CHANGE"
 export const CONTENT_TYPE_CHANGE = "CONTENT_TYPE_CHANGE"
 export const PAGINATION_CHANGE = "PAGINATION_CHANGE"
-export const UPDATE_SUCCESS_MESSAGE = "INTERCEPTION IS SUCCESSFULL"
+export const UPDATE_MESSAGE = "UPDATE_MESSAGE"
+export const UPDATE_INTERCEPTOR_STATUS = "UPDATE_INTERCEPTOR_STATUS"
```
revathskumar: Why a new action for updating the message? 
can't we use `UPDATE_MESSAGE` itself.

amit-mb: `UPDATE_MESSAGE` action is used in content script to send out a message of either `Intercept Success' or `Intercept Disabled`. `UPDATE_INTERCEPTOR_STATUS` action on other hand is used to update the switch button state in `UI`. Since we are seperating the switch toogle action from `content script's` success or failure, we need two actions

PR #103, opened 2018-04-11
Add logo image for app disabled state

cgirish:


cgirish:
screenshot for reference -
disabled-icon

app/background/background.ts
```diff
@@ -62,10 +62,30 @@ class BackgroundWorker {
...
+  updateBadgeIcon = (tabId:number, disabledStatus:boolean) => {
+    disabledStatus
+    ?
+      chrome.browserAction.setIcon({
+        path : {
+          "16" : "images/icon-disabled-16.png",
+      }, tabId
+      })
+    :
+      chrome.browserAction.setIcon({
```
revathskumar: duplication can be avoided. use the condition to determine the url only

Issue #53, opened 2018-03-19
Prefill the response into the text area


revathskumar:
Record the actual response and store in redux store. 

* The actual response should be in two different redux keys to enable restore option in future.
* The user will edit the json in text area.

PR #109, opened 2018-04-24
Fill textarea with response from backend

amit-mb:
#53


app/actions/index.ts
```diff
@@ -58,4 +61,22 @@ export function sendMessageToUI(message:string){
...
   return {type : UPDATE_INTERCEPTOR_STATUS, payload : {tabId, value}}
+}
+export function fetchResponse(url:string,method:string) {
+  return function (dispatch) {
+    axios({
+      method: method,
+      url: url
+    })
+    .then(response => response)
+      .then(response => dispatch(fetchSuccess(response)
```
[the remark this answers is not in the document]
revathskumar: If we are storing only response, the use `response.data`. 
But I think we need to store the response type as well

app/actions/index.ts
```diff
@@ -58,4 +61,20 @@ export function sendMessageToUI(message:string){
...
 export function updateInterceptorStatus(tabId:number, value:boolean){
   return {type : UPDATE_INTERCEPTOR_STATUS, payload : {tabId, value}}
+}
+export function fetchResponse(url:string,method:string, tabId:number, index: number) {
+  return function (dispatch) {
+    axios({
+      method: method,
+      url: url
+    })
+    .then(response => dispatch(response))
```
revathskumar: we can't just `dispatch`? where is the action?

app/actions/index.ts
```diff
@@ -58,4 +61,20 @@ export function sendMessageToUI(message:string){
...
+    .catch(err => dispatch(err))
```
revathskumar: action missing

app/reducers/rootReducer.ts
```diff
@@ -61,14 +61,34 @@ export const reducer = (state = INITIAL_POPUP_STATE, action: Action) => {
...
+          }
+        }
+    }
+  }
+  case actionType.FETCH_DATA_FAILURE: {
+    return {...state, responseData : {
+      ...state.responseData,
+      [action.payload.tabId]: {
+        ...state.responseData[action.payload.tabId],
+        [action.payload.index]: action.payload.error
```
revathskumar: why we are setting error in response data?

app/reducers/rootReducer.ts
```diff
@@ -61,14 +61,34 @@ export const reducer = (state = INITIAL_POPUP_STATE, action: Action) => {
...
+    case actionType.UPDATE_INTERCEPTOR_STATUS:
       return {
         ...state,
         isInterceptorOn: {...state.isInterceptorOn, [action.payload.tabId]: action.payload.value}
       };
+    case actionType.FETCH_DATA_SUCCESS: {
+        return {...state, responseData : {
+          ...state.responseData,
+          [action.payload.tabId]: {
+            ...state.responseData[action.payload.tabId],
```
revathskumar: this is suppose to be `requestId`

app/actions/index.ts
```diff
@@ -58,4 +61,20 @@ export function sendMessageToUI(message:string){
 }
 export function updateInterceptorStatus(tabId:number, value:boolean){
   return {type : UPDATE_INTERCEPTOR_STATUS, payload : {tabId, value}}
+}
+export function fetchResponse(url:string,method:string, tabId:number, index: number) {
+  return function (dispatch) {
+    axios({
+      method: method,
+      url: url
```
revathskumar: You need to pass the headers as well.

app/actions/index.ts
```diff
@@ -58,4 +61,20 @@ export function sendMessageToUI(message:string){
 }
 export function updateInterceptorStatus(tabId:number, value:boolean){
   return {type : UPDATE_INTERCEPTOR_STATUS, payload : {tabId, value}}
+}
+export function fetchResponse(url:string,method:string, tabId:number, index: number) {
+  return function (dispatch) {
+    axios({
```
revathskumar: return the promise

app/components/Intercept_Components/InterceptTextBox.tsx
```diff
@@ -8,65 +8,116 @@ export const InterceptTextBox = props => {
...
+            textAreaValue = JSON.stringify(props.responseData[props.rowProps.checkbox.requestId]) ||responseTextValue
+          })
+        }
+        }>X</span>
+          <label htmlFor="">Response Text</label>
+                <textarea
+                name="responseText"
+                className="responseText"
+                defaultValue={textAreaValue }
+                key={textAreaValue}
```
amit-mb: Added key attr to force re-render of `textarea` if changing defaultValue after clicking the `fetchResponse (X)` button

README.md
```diff
@@ -44,8 +44,13 @@ Henceforth the same AJAX request is requested by the browser, the browser is giv
 
 You can also stop listening for `AJAX calls` by clicking the `STOP LISTENING`
 
-The toggle switch is used to disable `INTERCEPTOR`. If the toggle is switched to `OFF` state,
-the extension will not mock any previously intercepted calls.
+The toggle switch is used to disable `INTERCEPTOR`. If the toggle is switched to `OFF` state, it displays a message saying `Interception Disabled` as below.
+
+<img src="images/interceptor_disabled.png" alt="Message shown by Interceptor on disabling">
```
revathskumar: these are already on master why this is showing as change here?

app/actions/index.ts
```diff
@@ -58,4 +62,16 @@ export function sendMessageToUI(message:string){
 }
 export function updateInterceptorStatus(tabId:number, value:boolean){
   return {type : UPDATE_INTERCEPTOR_STATUS, payload : {tabId, value}}
-}
\ No newline at end of file
+}
+export function fetchResponse(url:string,method:string, requestId:number) {
+  return {
+    type: FETCH_DATA,
```
revathskumar: this is suppose to be a async action right?

app/components/request_list.tsx
```diff
@@ -22,6 +22,7 @@ export interface RequestObj {
   disableInterceptor:React.ChangeEvent<HTMLButtonElement>;
   updateInterceptorStatus:React.ChangeEvent<HTMLButtonElement>;
   isInterceptorOn:object;
+  fetchResponse:React.MouseEvent<HTMLButtonElement>
```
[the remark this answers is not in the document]
amit-mb: Since `fetchResponse` is defined on `onClick` event, it is supposed to be` React.MouseEvent` i feel. Ref :https://reactjs.org/docs/events.html#mouse-events

app/components/Intercept_Components/InterceptTextBox.tsx
```diff
@@ -18,12 +20,22 @@ export const InterceptTextBox = props => {
         </a>
       </div>
       <div className="response">
-        <label htmlFor="">Response Text</label>
+        <label className="responseTextlabel">Response Text</label>
+        <span
+          className="fetch-responsetext"
+          onClick={() => {
+            props.fetchResponse(props.rowProps.checkbox);
+            textAreaValue = JSON.stringify(props.responseData[props.rowProps.checkbox.requestId]) || responseTextValue;
```
revathskumar: this shouldn't be there. no need for `|| responseTextValue`. thats should be handled in store.

app/store/aliases.ts
```diff
@@ -0,0 +1,27 @@
...
+        dispatch(fetchSuccess(data, requestId));
+      })
+      .catch(err => {
+        dispatch(fetchFailure(err, requestId));
+      });
+  };
+};
+
+export default {
+  FETCH_DATA: fetchDataAlias
```
revathskumar: please don't do this. 
please `export default fetchDataAlias`

app/components/Intercept_Components/InterceptTextBox.tsx
amit-mb: `responseTextValue` holds the value that is propagated from `onChange event` of `textarea`. If there is nothing typed in the textarea, Then it holds a value of empty string. Here `textAreaValue` holds the value that is propagated from the store as `props` when the `X(fetch response)` button is clicked. We could handle this in a `if else` conditional, but not in a store I guess. Any other approach you can think of. Please refer : https://github.com/code-mancers/interceptor/blob/18a38363d3f4f4700cc6c6e40fa25f875f528d2d/app/components/Intercept_Components/InterceptTextBox.tsx

app/store/aliases.ts
revathskumar: This is confusing can we use the normal thing since we are not integrating anything with background.

app/components/Intercept_Components/InterceptTextBox.tsx
revathskumar: okey I understand the `|| responseTextvalue` part. but still this statement should not be inside the `onClick.` This statement should be in the `render` method.

package.json
```diff
@@ -44,6 +44,7 @@
     "test:watch": "jest --watch"
   },
   "dependencies": {
+    "@types/axios": "^0.14.0",
```
kgrz: I'm not sure why these got updated, may be deliberate choice, but if not, ensure these changes don't happen. These can be caused due to running `npm upgrade` instead of `npm install` (or yarn equivalent)

app/store/aliases.ts
```diff
@@ -0,0 +1,28 @@
...
+    axios({
+      method,
+      url,
+      requestHeadersObject
+    })
+      .then(data => {
+        dispatch(fetchSuccess(data, requestId));
+        dispatch(handleRespTextChange(data.data, requestId))
+      })
+      .catch(err => {
```
kgrz: try to avoid `catch` when not required. The second argument to `then` here has the same affect as what catch does. using a `.catch` handler would mean that any JS errors or other errors also get caught. Not a huge problem here, but that can become a habit

app/store/aliases.ts
```diff
@@ -0,0 +1,28 @@
+import * as axios from "axios";
+import {fetchSuccess, fetchFailure, handleRespTextChange} from "./../actions";
+
+const fetchDataAlias = (payload: object) => {
+  return dispatch => {
+    const {method, url, requestId, requestHeaders} = payload.requestDetails;
+    const requestHeadersObject = requestHeaders.reduce((accumulatedObj, reqHeaderNameValuePair) => {
```
kgrz: can you explain what's the intent of this?

app/components/Intercept_Components/InterceptTextBox.tsx
```diff
@@ -2,12 +2,19 @@ import * as React from "react";
 import {RequestHeaderList} from "./RequestHeaderList";
 
 export const InterceptTextBox = props => {
+  const isObject = (val:any) => {
```
kgrz: A simpler check for this would be:

```javascript
return !value && typeof val === 'object';
```

So that would return true only if the value is truthy, and it's a type of object. `null` is also an object, but that's covered in the first check. stringify of a function is pretty much a noop, so that check is not required, we can treat it as not an object.

screenshot 2018-05-04 14 08 39

app/components/Intercept_Components/InterceptTextBox.tsx
```diff
@@ -2,12 +2,19 @@ import * as React from "react";
...
+  const requestId = props.rowProps.checkbox.requestId
+  console.log("Re-rendered")
+  console.log("responseData::", props.responseData[requestId])
   const defaultResponseText = "";
   const defaultStatusCode = "200";
   const defaultContentType = "application/json";
-  const responseTextValue = props.responseText[props.rowProps.checkbox.requestId] || defaultResponseText;
-  const statusCodeValue = props.statusCodes[props.rowProps.checkbox.requestId] || defaultStatusCode;
-  const contentTypeValue = props.contentType[props.rowProps.checkbox.requestId] || defaultContentType;
+  const responseTextValue = isObject(props.responseText[requestId]) ? JSON.stringify(props.responseText[requestId]) : props.responseText[requestId] || defaultResponseText;
```
kgrz: Ideally, it's more cleaner and easy to follow if the type of data is consistent: either always ensure it's an object, or ensure it's always a string (at this function call site). let the caller do the processing. That way, your main interception code doesn't need to handle all kinds of complex marshalling logic.

app/components/Intercept_Components/InterceptTextBox.tsx
```diff
@@ -18,13 +25,27 @@ export const InterceptTextBox = props => {
         </a>
       </div>
       <div className="response">
-        <label htmlFor="">Response Text</label>
+        <label className="responseTextlabel">Response Text</label>
+        <span
+          title="Fetch Response Text"
```
kgrz: why not use a `<a>` or a `button` tag for this? Also, title here doesn't seem to suggest what the action does

app/store/popup_store.ts
```diff
@@ -1,6 +1,9 @@
-import { createStore, applyMiddleware, compose } from 'redux'
+import { createStore, applyMiddleware } from 'redux'
 import { createLogger } from 'redux-logger'
 import { Middleware } from 'react-redux/node_modules/redux';
+import thunkMiddleware from 'redux-thunk';
+import aliases from './aliases'
+import { alias } from 'react-chrome-redux';
```
kgrz: what does this do?

app/store/aliases.ts
amit-mb: In axios, if we need to pass the request headers while making a request, we need to specify it this way `headers: {'X-Requested-With': 'XMLHttpRequest'},`.  But from the background script, we get the headers in the following format.
```js
requestHeaders : {
[0] : {name : 'X-Requested-With' , value : 'XMLHttpRequest'},
[1] : ...
}
So the function returns the headers in the way that axios requires.

Ref: https://github.com/axios/axios#request-config

File: app/components/Intercept_Components/InterceptTextBox.tsx (hunk 19 above)
Comment (amit-mb) [reply app/components/Intercept_Components/InterceptTextBox.tsx]: Yeah, wanted that to replace that with an icon before the release. Just put up a placeholder `span` tag for now

File: app/store/aliases.ts (hunk 15 above)
Comment (amit-mb) [reply app/store/aliases.ts]: The `catch` block here is executed in case the `axios` lib isn't able to complete and fetch the response.  Don't you recoemmend using this in case the server returns a error such as `404` or `500`. Refer [this](https://github.com/axios/axios#handling-errors)

File: app/store/popup_store.ts (hunk 20 above)
Comment (amit-mb) [reply app/store/popup_store.ts]: Refer [this](https://github.com/tshaddix/react-chrome-redux/wiki/Advanced-Usage#alias)

Comment (kgrz) [reply app/store/popup_store.ts]: Okay, more precisely, why do you want to use this?

Comment (kgrz) [reply app/store/popup_store.ts]: nevermind, I think I understand.

File: app/components/Intercept_Components/InterceptTextBox.tsx [hunk 21]
```diff
@@ -2,12 +2,13 @@ import * as React from "react";
...
 
 export const InterceptTextBox = props => {
+  const requestId = props.rowProps.checkbox.requestId
   const defaultResponseText = "";
   const defaultStatusCode = "200";
   const defaultContentType = "application/json";
-  const responseTextValue = props.responseText[props.rowProps.checkbox.requestId] || defaultResponseText;
-  const statusCodeValue = props.statusCodes[props.rowProps.checkbox.requestId] || defaultStatusCode;
-  const contentTypeValue = props.contentType[props.rowProps.checkbox.requestId] || defaultContentType;
+  let responseTextValue = props.responseText[requestId] || defaultResponseText;
```
revathskumar: why `let`. can be `const`

app/components/Intercept_Components/InterceptTextBox.tsx
```diff
@@ -18,13 +19,26 @@ export const InterceptTextBox = props => {
...
       </div>
       <div className="response">
-        <label htmlFor="">Response Text</label>
+        <label className="responseTextlabel">Response Text</label>
+        <button
+          className="fetch-responsetext btn-sm btn-primary"
+          onClick={() => {
+            props.fetchResponse(props.rowProps.checkbox)
+        }
+      }
```
revathskumar: fix indentation.

app/store/aliases.ts
```diff
@@ -0,0 +1,29 @@
...
+      url,
+      requestHeadersObject
+    })
+      .then(data => {
+        dispatch(fetchSuccess("", requestId))
+        dispatch(handleRespTextChange(JSON.stringify(data.data), requestId))
+        dispatch(fetchSuccess(JSON.stringify(data.data), requestId));
+      })
+      .catch(err => {
+        dispatch(fetchFailure("", requestId))
```
revathskumar: error message can't be blank.

app/store/aliases.ts
```diff
@@ -0,0 +1,29 @@
...
+    }, {});
+    axios({
+      method,
+      url,
+      requestHeadersObject
+    })
+      .then(data => {
+        dispatch(fetchSuccess("", requestId))
+        dispatch(handleRespTextChange(JSON.stringify(data.data), requestId))
+        dispatch(fetchSuccess(JSON.stringify(data.data), requestId));
```
revathskumar: avoid repeated calls of `JSON.stringify`

Issue #123, opened 2018-05-16
Add tooltips to all buttons


amit-mb:
Add tool tips to all buttons describing what they do

amit-mb:
Closing this. #126

PR #126, opened 2018-05-16
Add title attribute to all buttons

amit-mb:
#123

### DIFF
--- a/app/containers/popup.tsx
+++ b/app/containers/popup.tsx
@@ -132,7 +132,6 @@ export class Popup extends React.Component<POPUP_PROPS & DispatchProps, {}> {
       "button-stop-listening btn-danger": this.props.enabled
     });
     const props = this.props;
-    const enabled = this.props.enabled;
     return (
       <div className="popup">
         <header>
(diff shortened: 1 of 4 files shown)


app/containers/popup.tsx
```diff
@@ -164,7 +164,12 @@ export class Popup extends React.Component<POPUP_PROPS & DispatchProps, {}> {
               </svg>
               <span>INTERCEPTOR</span>
             </a>
-            <button type="button" onClick={this.handleClick} className={buttonClass}>
+            <button
+              title="Start/Stop Listening to Requests"
```
revathskumar: can be conditionally set either start or stop.

app/containers/popup.tsx
```diff
@@ -164,45 +160,45 @@ export class Popup extends React.Component<POPUP_PROPS & DispatchProps, {}> {
               </svg>
               <span>INTERCEPTOR</span>
             </a>
-            <button type="button" onClick={this.handleClick} className={buttonClass}>
-              {this.props.enabled ? "Stop Listening" : "Start Listening"}
+            <button
+              title={enabled ? "Stop Listening to Requests" : "Start Listening to Requests"}
```
revathskumar: use `props.enabled` to make it consistent across the component. No need for `enabled` as separate const.

PR #127, opened 2018-05-16
Fix typescript errors

amit-mb:
Fix some typescript errors and prettified code


app/content/content.ts
```diff
@@ -24,69 +36,77 @@ class Intercept {
...
+    while (document.querySelectorAll(querySelector).length) {
+      let elemToRemove = document.querySelector(querySelector);
+      elemToRemove.parentNode.removeChild(elemToRemove);
     }
-  }
+  };
 
-  runInterceptor  = (selectedReqs) => {
+  runInterceptor = (selectedReqs: selectedReqs) => {
+    console.log(selectedReqs);
```
revathskumar: can be removed

app/types/index.ts
```diff
@@ -1,52 +1,53 @@
...
+  method?: string;
+  url?: string;
+  requests: Array<any>;
+  errorMessage: string;
+  checkedReqs: object;
+  statusCodes: object;
+  responseText: object;
+  contentType: object;
+  PageDetails: Array<object>;
+  interceptStatus: string;
```
revathskumar: instead of `interceptSatus` as `string`. it can be union/enum state. 
https://www.typescriptlang.org/docs/handbook/enums.html

app/.prettierrc
```diff
@@ -0,0 +1,5 @@
+{
+  "printWidth": 100,
+  "parser": "typescript",
+  "tabWidth": 2
```
revathskumar: Is there any way we can say indent with `spaces`?

amit-mb: We can specify `"useTabs": false` for preventing indentation with tabs. But that is `false` by default. So no need. See [this](https://prettier.io/docs/en/options.html#tabs)

PR #132, opened 2018-05-22
Refactor redux store 

amit-mb:
Entire `redux` store has been restructured in the following form(in its initial state) making it clear to make further updates as opposed to a messy structure earlier :
```js
store : {
  currentTab: tabId,
  interceptStatus: "",
  currentUrl: "",
    tabRecord : {
      tabId: {
        PageDetails: [],
        checkedReqs: {},
        enabledStatus: false,
        errorMessage: "",
        isInterceptorOn: true,
        requests: [],
       },
    requestRecords : {
    [requestId] : {
    statusCodes : {},
    contentType : {},
    responseData : {},
    responseText : {},
    responseData : {}
    }
  }
}
```


app/__tests__/intercept_box.test.jsx
```diff
@@ -1,21 +1,34 @@
...
-import {shallow} from "enzyme";
-import {InterceptTextBox} from "./../components/Intercept_Components/InterceptTextBox";
+import { shallow } from "enzyme";
+import { InterceptTextBox } from "./../components/Intercept_Components/InterceptTextBox";
 
 const createTestProps = (props = {}) => ({
   // common props
-  handleIntercept: jest.fn(),
+  currentTabId: 2347,
+  data: {
```
kgrz: consider naming this something other than `data`

app/containers/popup.tsx
```diff
@@ -30,73 +26,59 @@ const isChromeUrl = (url: string) => {
...
   clearRequests = (_: React.MouseEvent<HTMLButtonElement>): void => {
-    MessageService.clearData(this.props.tabId);
-    this.props.clearFields();
+    this.props
+      .clearFields(this.props.currentTab)
+      .then(() => {
+        MessageService.setBadgeTextToZero(this.props.currentTab);
+      })
+      .catch((err: any) => {
+        console.log(err);
```
kgrz: can the log statement be removed? Or may be remove those before shipping the final bundle.

app/containers/popup.tsx
```diff
@@ -30,73 +26,59 @@ const isChromeUrl = (url: string) => {
...
+      });
   };
 
-  handleCheckToggle = (reqId: number, presentCheckedState: boolean): void => {
-    this.props.handleCheckToggle(reqId, presentCheckedState);
+  handleCheckToggle = (tabId: number, reqId: number, presentCheckedState: boolean): void => {
```
kgrz: where is tabId being used?

app/containers/popup.tsx
```diff
@@ -30,73 +26,59 @@ const isChromeUrl = (url: string) => {
...
+    this.props.handleCheckToggle(this.props.currentTab, reqId, presentCheckedState);
   };
 
-  handleRespTextChange = (value: string, requestId: string): void => {
-    this.props.handleRespTextChange(value, requestId);
+  handleRespTextChange = (value: string, requestId: string, tabId: number): void => {
```
kgrz: I feel this indirection can be avoided. Any reason not to call the `this.props.handle...` function wherever it's used?

app/containers/popup.tsx
```diff
@@ -108,29 +90,32 @@ export class Popup extends React.Component<POPUP_PROPS & DispatchProps, {}> {
...
-      this.updateBadgeIcon(this.props.tabId, false);
+      this.props.updateInterceptorStatus(this.props.currentTab, true);
+      this.updateBadgeIcon(this.props.currentTab, false);
     }
   };
 
   render() {
+    const props = this.props;
+    if (!props.data) {
+      return null;
```
kgrz: does this mean we end up showing a blank window under this condition?

app/background/background.ts
```diff
@@ -80,32 +76,36 @@ class BackgroundWorker {
     });
   };
 
-  updateBadgeText = (noOfRequests: number) => {
+  updateBadgeText = (tabId: number, requestsLength: number) => {
     chrome.browserAction.setBadgeText({
-      text: `${noOfRequests}`,
-      tabId: this.currentTab
+      text: `${requestsLength}`,
+      tabId: tabId
```
revathskumar: Use shorthand.
```
{
  text: `${requestsLength}`,
   tabId
}
```

app/components/Intercept_Components/InterceptTextBox.tsx
```diff
@@ -1,33 +1,30 @@
...
   const requestId = props.rowProps.checkbox.requestId;
   const defaultResponseText = "";
   const defaultStatusCode = "200";
   const defaultContentType = "application/json";
-  const responseTextValue = props.responseText[requestId] || defaultResponseText;
-  const statusCodeValue = props.statusCodes[requestId] || defaultStatusCode;
-  const contentTypeValue = props.contentType[requestId] || defaultContentType;
+  const responseTextValue = props.data.responseText[requestId] || defaultResponseText;
+  const statusCodeValue = props.data.statusCodes[requestId] || defaultStatusCode;
+  const contentTypeValue = props.data.contentType[requestId] || defaultContentType;
```
revathskumar: All three default values should go in `initialTabProperties` in reducer.

app/components/Intercept_Components/index.tsx
```diff
@@ -1,23 +1,18 @@
...
+import { InterceptTextBox } from "./InterceptTextBox";
+import { RequestObj } from "./../request_list";
 
-
-export const InterceptForm:React.SFC<RequestObj> = (props) => {
+export const InterceptForm: React.SFC<RequestObj> = props => {
   return (
     <div>
       <InterceptTextBox
+        data={props.data}
```
revathskumar: Why this component need whole data? This is per row and can be should accept only the fields it need for that particular row.

app/components/Switch.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from "react";
 
-export const Switch = function(props = {isOn: true, handleSwitch: () => {}}) {
+export const Switch = (props = { isOn: true, handleSwitch: () => {} }) => {
```
revathskumar: Don't default this way. Use `defaultProps`. Also no need to default `handleSwitch` with empty method here.
make it as required method.

app/components/request_list.tsx
```diff
@@ -4,36 +4,27 @@ import matchSorter from "match-sorter";
 import { InterceptForm } from "./../components/Intercept_Components/index";
 import { InterceptAllButton } from "./../components/InterceptAllButton";
 import { Switch } from "./Switch";
-import { responseField, statusCodes, contentType } from "./../content/content";
-import { interceptOn } from "./../types";
+// import { responseField, statusCodes, contentType } from "./../content/content";
+// import { interceptOn } from "./../types";
```
revathskumar: can be removed

app/components/request_list.tsx
```diff
@@ -4,36 +4,27 @@ import matchSorter from "match-sorter";
...
-  disableInterceptor: (tabId:number) => void ;
-  updateInterceptorStatus: (tabId:number) => void;
-  isInterceptorOn: interceptOn;
+  handlePaginationChange?: (rowSize: number, tabId: number, field: string) => void;
+  clearRequests?: (e: React.MouseEvent<HTMLButtonElement>) => void;
+  updateInterceptorStatus?: (tabId: number) => void;
   fetchResponse: React.MouseEvent<HTMLSpanElement>;
-  responseData: object;
-  responseError: object;
+  handleSwitch: any;
```
revathskumar: this is suppose to have function signature. not any

app/components/request_list.tsx
```diff
@@ -4,36 +4,27 @@ import matchSorter from "match-sorter";
 import { InterceptForm } from "./../components/Intercept_Components/index";
```
revathskumar: Rename the file to `RequestList.tsx`

app/components/request_list.tsx
```diff
@@ -123,36 +115,30 @@ const RequestList: React.SFC<RequestObj> = props => {
...
         showPagination={true}
         showPaginationTop={false}
         showPaginationBottom={true}
         defaultPageSize={10}
-        page={props.PageDetails[props.tabId] ? props.PageDetails[props.tabId].currentPageNumber : 0}
-        pageSize={
-          props.PageDetails[props.tabId] ? props.PageDetails[props.tabId].currentRowSize : 10
-        }
+        page={props.data.PageDetails ? props.data.PageDetails.currentPageNumber : 0}
+        pageSize={props.data.PageDetails ? props.data.PageDetails.currentRowSize : 10}
```
revathskumar: Please default this in `initialTabProperties`.

app/components/request_list.tsx
```diff
@@ -123,36 +115,30 @@ const RequestList: React.SFC<RequestObj> = props => {
...
+          props.handlePaginationChange(changedPageNo, props.currentTabId, "currentPageNumber")
         }
         onPageSizeChange={changedRowSize =>
-          props.handlePaginationChange(changedRowSize, props.tabId, "currentRowSize")
+          props.handlePaginationChange(changedRowSize, props.currentTabId, "currentRowSize")
         }
         collapseOnDataChange={false}
         SubComponent={row => (
           <InterceptForm
+            data={props.data}
```
revathskumar: Why this need whole data? pass only the data for that request.

app/containers/popup.tsx
```diff
@@ -30,73 +26,59 @@ const isChromeUrl = (url: string) => {
...
   isUrlInValid = (tabUrl: string) => {
     return !tabUrl || isChromeUrl(tabUrl);
   };
 
   handleClick = (_: React.MouseEvent<HTMLButtonElement>): void => {
-    if (this.isUrlInValid(this.props.tabUrl)) {
-      this.props.errorNotify(`Cannot Start Listening on ${this.props.tabUrl}`);
+    const props = this.props;
+    if (this.isUrlInValid(this.props.currentUrl)) {
+      props.errorNotify(`Cannot Start Listening on ${this.props.currentUrl}`, props.currentTab);
```
revathskumar: consistently use `this.props` or `props`.
If we decide to use `this.props` in whole method, remove the unwanted const props.

app/content/content.ts
```diff
@@ -41,13 +41,19 @@ class Intercept {
...
-    const checkedReqs = presentState.requests.filter( (req:chrome.webRequest.WebRequestBodyDetails) => {
-      return presentState.checkedReqs[req.requestId] && tabId;
-    });
+    const presentState = this.store.getState().data[tabId];
+    let checkedReqs;
+    if (presentState) {
+      checkedReqs = presentState.requests.filter((req: chrome.webRequest.WebRequestBodyDetails) => {
+        return presentState.checkedReqs[req.requestId] && tabId;
+      });
+    } else {
```
revathskumar: Why this else block?

app/reducers/rootReducer.ts
```diff
@@ -1,94 +1,140 @@
...
   responseData: {},
-  responseError: {}
+  responseError: {},
+  responseText: {},
+  statusCodes: {}
 };
 
-//ACTION CONSTANTS
-import * as actionType from "../actions";
+function extendStateData(state: POPUP_PROPS, payload: any, props: any) {
```
revathskumar: what are `props` here?

app/reducers/rootReducer.ts
```diff
@@ -1,94 +1,140 @@
...
+      return extendStateData(state, action.payload, {
+        requests: []
+      });
     case actionType.TOGGLE_CHECKBOX:
-      return {
-        ...state,
+      return extendStateData(state, action.payload, {
         checkedReqs: {
-          ...state.checkedReqs,
+          ...state.data[action.payload.tabId].checkedReqs,
```
revathskumar: Why do we need to pass this again frmo here. This is already doing in `extendStateData`.

app/reducers/rootReducer.ts
```diff
@@ -1,94 +1,140 @@
...
 };
 
-//ACTION CONSTANTS
-import * as actionType from "../actions";
+function extendStateData(state: POPUP_PROPS, payload: any, props: any) {
+  return {
+    ...state,
+    data: {
+      ...state.data,
+      [payload.tabId]: {
```
revathskumar: I think here we using only the `tabId` from payload.
But we can simplify this by avoiding the `props` and using the `payload` alone.

```
const { tabId, ...otherPayloadItems } = payload
return {
    ...state,
    data: {
      ...state.data,
      [tabId]: {
        ...state.data[tabId],
        ...otherPayloadItems,
     }
  }
}
```

This way we can avoid the whole `props` argument.

app/background/background.ts
```diff
@@ -37,29 +46,16 @@ class BackgroundWorker {
         };
       }
       switch (request.message) {
-        case "CLEAR_DATA": {
-          this.clearData();
+        case "SET_BADGE_ZERO": {
```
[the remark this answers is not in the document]
amit-mb: It is a generic action. Since we are using the messaging call to set it specifically to`0`, I had changed it. Using generic action names makes sense :thumbsup: will revert it

app/containers/popup.tsx
amit-mb: Its being used in the reducer to add a `[requestId] : boolean` property in the store in form of `store.data.[tabId].checkedReqs.[requestId] : true || false`.

app/containers/popup.tsx
amit-mb: Yes, we dispatch an action with `tabId` as an argument. Whenever we open the `popup`, It takes some time(less than 2 sec) to reflect the `data.[tabId]` value in the store and ends up giving an error if this condition wasn't there. Until then, we show a blank window. And it only happens for the first time after the extension is freshly installed.

app/content/content.ts
revathskumar: then the condition should be return as 

```
if(!presentState) {
  return;
}

// rest of the things
```

app/containers/popup.tsx
kgrz: You're not passing this variable into the function that's getting called. Check once?

app/containers/popup.tsx
kgrz: okay. Try adding a loader or something (not a gif, but something like what FB does when loading the home page, for instance. May be in a later PR.

app/containers/popup.tsx
```diff
@@ -159,81 +144,65 @@ export class Popup extends React.Component<POPUP_PROPS & DispatchProps, {}> {
...
+                this.props.data.enabledStatus
+                  ? "Stop Listening to Requests"
+                  : "Start Listening to Requests"
+              }
               type="button"
               onClick={this.handleClick}
               className={buttonClass}
             >
-              {props.enabled ? "Stop Listening" : "Start Listening"}
+              {this.props.data.enabledStatus ? "Stop Listening" : "Start Listening"}
```
kgrz: move this to a function? reuse that here and in the title

app/reducers/rootReducer.ts
amit-mb: This can be done. But this only does a shallow merging of objects and not deep merging. This works for actions such as `UPDATE_REQUESTS` and `TOGGLE_LISTENING`, but we need deep merging of objects in actions like `RESP_TEXT_CHANGE` and `TOGGLE_CHECKBOX`. Any other way we could achieve that?

revathskumar: Only for those two do the deep merge in reducer and send the new object as payload to this method.

app/components/Intercept_Components/InterceptTextBox.tsx
amit-mb: Since `requestId` and `tabId` are dynamic values that get saved in the store after dispatching an `INITIALISE_DEFAULTS` action , how do I add these in the reducer without the values of `tabId` or `requestId`

revathskumar:
@amit-mb 

After testing some issues found.
* The message `Interception success` never gets removed. Even after we close and reopen the popup
  Also show `interception success` message in a good format with background color of green or   something.
* After clear, it clears only the redux store, means after clear then again record another request the old cleared requests will comes back.

app/components/RequestList.tsx
```diff
@@ -123,36 +113,30 @@ const RequestList: React.SFC<RequestObj> = props => {
...
+          props.handlePaginationChange(changedPageNo, props.currentTabId, "currentPageNumber")
         }
         onPageSizeChange={changedRowSize =>
-          props.handlePaginationChange(changedRowSize, props.tabId, "currentRowSize")
+          props.handlePaginationChange(changedRowSize, props.currentTabId, "currentRowSize")
         }
         collapseOnDataChange={false}
         SubComponent={row => (
           <InterceptForm
+            tabRecord={props.tabRecord}
```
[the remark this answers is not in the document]
amit-mb: The structure of the store is as follows : 
```javascript
store : {
  currentTab: tabId,
  currentUrl: "",
    tabRecord : {
      tabId: {
        PageDetails: [],
        checkedReqs: {},
        contentType: {},
        enabledStatus: false,
        errorMessage: "",
        interceptStatus: "",
        isInterceptorOn: true,
        requests: [],
        responseData: {},
        responseError: {},
        responseText: {},
        statusCodes: {}
       }
    }
}
```

The `InterceptForm` component needs the `statusCodes`, `contentType`,  `responseData`, `responseError` and `responseText` props from the parent. 

The store has to be re-structured  isthe following way to achieve this  : 
```js
store : {
  currentTab: tabId,
  currentUrl: "",
    tabRecord : {
      tabId: {
        PageDetails: [],
        checkedReqs: {},
        enabledStatus: false,
        errorMessage: "",
        interceptStatus: "",
        isInterceptorOn: true,
        requests: [],
       },
    requestRecords : {
    [requestId] : {
    statusCodes : {},
    contentType : {},
    responseData : {},
    responseText : {},
    responseData : {}
    }
  }
}
```

Here, we can pass just the required props as follows:

```html
<InterceptForm requestRecord={props.requestRecords[row.requestId]} ... />
```
This is how it needs to be done right?

app/components/RequestList.tsx
```diff
@@ -123,36 +113,30 @@ const RequestList: React.SFC<RequestObj> = props => {
...
+            currentTabId={props.currentTabId}
             rowProps={row}
```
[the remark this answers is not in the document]
amit-mb: `row` contains the `url`, `requestId`and `requestHeaders` for the current `row`

app/components/RequestList.tsx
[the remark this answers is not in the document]
revathskumar: Thats a good approach on requestRecords. Keep these kinda suggestions coming. 
I missed this actually.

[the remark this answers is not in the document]
amit-mb: Also, if the store structure is refactored to include `requestRecords`, we could also convert `statusCodes`, `contentType`, `responseData` and `responseType` to string type instead of object.

app/components/RequestList.tsx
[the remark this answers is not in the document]
revathskumar: okey

app/__tests__/popup.test.jsx
```diff
@@ -1,27 +1,37 @@
...
+    contentType: {},
+    enabledStatus: false,
+    errorMessage: "",
+    interceptStatus: "",
+    isInterceptorOn: true,
+    requests: [],
+    responseData: {},
+    responseError: {},
+    responseText: {},
+    statusCodes: {}
```
revathskumar: Isn't this got changed to `requestRecords`?

app/__tests__/request_list.test.jsx
```diff
@@ -1,38 +1,48 @@
...
+    contentType: {},
+    enabledStatus: true,
+    errorMessage: "",
+    interceptStatus: "",
+    isInterceptorOn: true,
+    requests: [],
+    responseData: {},
+    responseError: {},
+    responseText: {},
+    statusCodes: {}
```
revathskumar: requestRecords?

package.json
```diff
@@ -52,8 +52,8 @@
     "url": "https://github.com/code-mancers/interceptor.git"
   },
   "scripts": {
-    "build": "webpack -p --config webpack.config.prod.js",
-    "watch": "webpack --config webpack.config.dev.js --watch --progress --colors",
+    "build": "./node_modules/.bin/webpack -p --config webpack.config.prod.js",
```
revathskumar: Why this change? npm script will find the executable from `./node_modules/.bin/`. no need to specify this.

amit-mb: Okay, will revert it back. Didn't know about this ^^

PR #133, opened 2018-06-15
Add url interception without listening

amit-mb:
- Adds ability to intercept requests without explicitly having to listen to them

Add URL Modal:

screen shot 2018-06-18 at 12 34 31 pm

After the URL is added for interception:

screen shot 2018-06-18 at 12 34 41 pm

screen shot 2018-06-18 at 12 34 55 pm

- Add ability to edit URL's inline by changing the URL fields 

edit_url

- more screenshots (Design by @cgirish )

screen shot 2018-07-13 at 12 37 06 pm

screen shot 2018-07-13 at 7 21 50 pm

screen shot 2018-07-13 at 7 22 01 pm

screen shot 2018-07-13 at 7 22 14 pm

screen shot 2018-07-13 at 7 22 41 pm


app/__tests__/popup.test.jsx
```diff
@@ -108,7 +108,12 @@ describe("Popup", () => {
         "Cannot Start Listening on chrome://version",
         1
       );
-      expect(wrapper.find("button").hasClass("button-start-listening")).toEqual(true);
+      expect(
+        wrapper
+          .find("button")
+          .at(0)
+          .hasClass("button-start-listening")
+      ).toEqual(true);
```
revathskumar: use `toBeTruthy()`

app/components/AddRuleModal.tsx
```diff
@@ -0,0 +1,36 @@
...
+export const AddRuleModal: React.SFC<{}> = props => {
+  return (
+    <Modal show={props.showModal} handleClose={props.handleClose}>
+      <label htmlFor="modal-url">URL</label>
+      <input
+        className="form-control"
+        type="text"
+        name="modal-url"
+        id="url-input-modal"
+        onChange={e => props.updateModalUrl(e.target.value)}
```
revathskumar: `onChange={props.updateModalUrl}`.
`updateModalUrl` will receive `event` object as first argument and do the `e.target.value` inside  `updateModalUrl`.

Advantage: avoiding function declaration on reach render.

app/components/AddRuleModal.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from "react";
+import { Modal } from "./ModalWrapper";
+export const AddRuleModal: React.SFC<{}> = props => {
+  return (
+    <Modal show={props.showModal} handleClose={props.handleClose}>
+      <label htmlFor="modal-url">URL</label>
```
revathskumar: `for` should be associated with `id` not name

app/components/AddRuleModal.tsx
```diff
@@ -0,0 +1,36 @@
...
+      <input
+        className="form-control"
+        type="text"
+        name="modal-url"
```
revathskumar: name like `modal-url` can be better. no need to have modal in context for the name of this input. 
better will be `request_url`.

avoid `-` (hypen) for names.

app/components/AddRuleModal.tsx
```diff
@@ -0,0 +1,36 @@
...
+        id="url-input-modal"
+        onChange={e => props.updateModalUrl(e.target.value)}
+      />
+      <label htmlFor="modal-method">Method</label>
```
revathskumar: `for` should be associated with `id` not name

app/components/AddRuleModal.tsx
```diff
@@ -0,0 +1,36 @@
...
+      <div>
+        <select
+          defaultValue="GET"
+          value={props.modalMethod}
+          className="modal-method"
+          onChange={e => props.updateModalMethod(e.target.value)}
```
revathskumar: `onChange={props.updateModalUrl}`. for the same reason above.

app/components/ModalWrapper/index.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from "react";
+export const Modal = ({ handleClose, show, children }) => {
+  const showHideClassName = show ? "modal display-block" : "modal display-none";
```
revathskumar: no need to switch the css classes.
Instead either we render or no render the component.

app/containers/Popup.tsx
```diff
@@ -28,6 +31,11 @@ const isChromeUrl = (url: string) => {
   return CHROME_URL_REGEX.test(url);
 };
 export class Popup extends React.Component<POPUP_PROPS & DispatchProps, {}> {
+  state = {
```
revathskumar: move to redux.

app/components/AddRuleModal.tsx
amit-mb: Had missed this point^^. Thanks for reminding. Better to use this pattern everytime we pass event/event properties

app/containers/Popup.tsx
amit-mb: Since its a `container` component and the `state` defined here is used just to show/hide the `modal` window(not much complication regarding state involved), should we move this to redux ?

app/__tests__/AddRuleModal.test.jsx
```diff
@@ -0,0 +1,62 @@
...
+    wrapper = shallow(<AddRuleModal {...props} />);
+  });
+
+  describe("default", () => {
+    test("Contains a Modal HOC component", () => {
+      expect(wrapper).toBeDefined();
+      expect(wrapper.find("Modal")).toHaveLength(1);
+    });
+    test("Contains a input, select and a button element", () => {
+      expect(wrapper.find("input")).toBeTruthy();
```
revathskumar: AFAIK, this won't work. `wrapper.find("input")` always returns an object. so we have to check for the length to make sure the element exists

app/__tests__/AddRuleModal.test.jsx
```diff
@@ -0,0 +1,62 @@
+import * as React from "react";
+import { shallow } from "enzyme";
+import { AddRuleModal } from "./../components/AddRuleModal";
+
+const createTestProps = props => ({
+  handleClose: jest.fn(),
+  showModal: false,
+  modalMethod: "GET",
+  addRequest: jest.fn(),
+  modalUrl: "www.codemancers.com",
```
revathskumar: Don't modal data based on UI. what if tomorrow we decided to not go with `Modal`? 
the name become meaningless and since data is used everywhere, renaming will impact a lot of places.

revathskumar:
@cgirish Can you take a look into this for a better design? 
Should we really need to go with `Modal` any alternate approach? 


@amit-mb if we are deciding to go with `Modal` we need a backdrop.

app/actions/index.ts
```diff
@@ -69,5 +72,29 @@ export function initialiseDefaults(
   currentUrl: string,
   interceptStatus: string
 ) {
-  return { type: INITIALISE_DEFAULTS, payload: { currentTab, currentUrl, interceptStatus } };
+  return {
+    type: INITIALISE_DEFAULTS,
+    payload: { currentTab, currentUrl, interceptStatus }
```
revathskumar: Since we already have `initialState`, why we need to send the payload?

app/components/ModalWrapper/index.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from "react";
+export const Modal = ({ handleClose, show, children }) => {
+  if (!show) {
```
revathskumar: This condition shouldn't be a part of this component. 
```
{showModal && <Modal {...props} />}
```

app/actions/index.ts
amit-mb: Although `initialState` is defined in the reducer, we dont have any data about the `currentTabId`, `currentTabUrl`. Therefore, I dispatch this action as soon as the `popup` is opened so that we can store information based on `tabId`. Better to rename this as `currentTabId`, `currentTabUrl`.

I also dispatch `interceptStatus` with empty string so that after each time the popup is opened, the `Success` or `Error` message gets erased.

app/__tests__/AddRuleModal.test.jsx
amit-mb: Yes. ^^ that makes sense. Will change names.

revathskumar:
@cgirish @amit-mb Are we going forward with `Modal` itself? Or any better approach we are looking?

revathskumar:
@amit-mb Also please put the error message of URL validaiton near the text box itself. Don't put those error message in global error.

amit-mb:
I have also added the ability to edit url's in this PR (see 4th screenshot in PR description). Let's for @girish and check if there are any better design approaches.

@revathskumar , `error message of URL validaiton near the text box itself`,good idea than showing as global error, i will make necessary changes required and push 👍

app/components/AddRuleModal.tsx
```diff
@@ -0,0 +1,105 @@
...
+  addRequest: (url: string, method: string) => void;
+  addRequestMethod: string;
+  addRequestUrl: string;
+  addRuleErrorNotify: (errorMessage: string, tabId: number) => void;
+  handleClose: () => void;
+  tabId: number;
+  updateAddRequestMethod: (value: string, tabId: number) => void;
+  updateAddRequestUrl: (value: string, tabId: number) => void;
+}
+export const AddRuleModal: React.SFC<AddRuleModalProps> = props => {
```
revathskumar: Make it extend from `Component`. 
As already discussed, if we need instance methods never use `Function` component.

app/components/AddRuleModal.tsx
```diff
@@ -0,0 +1,105 @@
...
+          </select>
+        </div>
+      </div>
+      <div className="modal-footer text-right">
+        <button className="btn" onClick={handleClose}>
+          Cancel
+        </button>
+        <button
+          className="btn btn-primary btn-add-rule"
+          onClick={() => {
```
revathskumar: Avoid inline method and move to instance method.

app/components/AddRuleModal.tsx
```diff
@@ -0,0 +1,105 @@
...
+      new URL(str);
+      return true;
+    } catch {
+      return false;
+    }
+  };
+
+  const handleClose = () => {
+    //erase the previously set error message on each re-render
+    props.addRuleErrorNotify(``, props.tabId);
```
revathskumar: Why using backticks?

app/components/AddRuleModal.tsx
```diff
@@ -0,0 +1,105 @@
...
+    //reset the url to empty string and request method to "GET"
+    props.updateAddRequestUrl(``, props.tabId);
+    props.updateAddRequestMethod(`GET`, props.tabId);
```
revathskumar: backticks

app/components/AddRuleModal.tsx
```diff
@@ -0,0 +1,105 @@
...
+    } catch {
+      return false;
+    }
+  };
+
+  const handleClose = () => {
+    //erase the previously set error message on each re-render
+    props.addRuleErrorNotify(``, props.tabId);
+    //reset the url to empty string and request method to "GET"
+    props.updateAddRequestUrl(``, props.tabId);
```
revathskumar: backticks

app/components/RequestList.tsx
```diff
@@ -2,28 +2,24 @@ import * as React from "react";
...
   handleRespTextChange?: () => (value: string, reqId: string) => void;
   handleStatusCodeChange?: (value: string, reqId: string) => void;
   handleCheckToggle?: (tabId: number, reqId: number, checked: boolean) => void;
   handleContentTypeChange?: (value: string, reqId: string) => void;
   handlePaginationChange?: (rowSize: number, tabId: number, field: string) => void;
-  clearRequests?: (e: React.MouseEvent<HTMLButtonElement>) => void;
   updateInterceptorStatus?: (tabId: number) => void;
   fetchResponse: React.MouseEvent<HTMLSpanElement>;
   handleSwitchToggle: any;
+  handleChangeUrl: any;
```
revathskumar: duplicate?

app/components/AddRuleModal.tsx
amit-mb: We are using `functions`(defined internally) and not `methods` or `lifecycle methods` as well.  Do we really need a `stateful` component here ?

revathskumar: functions defining in render method is not good, this is similar to it. 
This will define all the methods again and again when each time it renders.

By using `Component` doesn't mean `stateful`. we are not using internal state.

PR #134, opened 2018-06-19
Add support for interception against page refreshes

amit-mb:
- Tested [here](http://output.jsbin.com/sumawo/), [here](https://output.jsbin.com/bulunet) and [here](https://t9-keyboard.herokuapp.com/)
- Does not work on some websites due to caching (?) Not sure

### DIFF
--- a/app/content/content.ts
+++ b/app/content/content.ts
@@ -167,7 +167,7 @@ class Intercept {
         }
       })();`;
     let script = document.createElement("script");
-    script.defer = true;
+    script.defer = false;
     script.id = "disableInterceptorScript";
     script.type = "text/javascript";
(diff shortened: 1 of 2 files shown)


app/content/content.ts
```diff
@@ -167,7 +167,7 @@ class Intercept {
         }
       })();`;
     let script = document.createElement("script");
-    script.defer = true;
+    script.defer = false;
```
amit-mb: I was having one concern about this change. Does making make everypage load `slower` since this script has be loaded and executed before rendering the page iteself ?

revathskumar: can you test this out and come up with some stats?

PR #135, opened 2018-06-28
Add some basic properties such as name and desc to package.json

amit-mb:
- Add some basic properties such as name and desc to package.json

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,4 +1,21 @@
 {
+  "name": "interceptor",
+  "version": "0.0.2",
+  "description":
+    "A browser extension that mocks AJAX request at the browser level so you can run front ends without really starting a backend server.",
+  "license": "MIT",
+  "homepage": "https://github.com/code-mancers/interceptor#readme",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/code-mancers/interceptor"
+  },
+  "keywords": [
+    "interceptor",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -1,4 +1,21 @@
 {
+  "name": "interceptor",
+  "version": "0.2.1",
+  "description":
+    "A browser extension that mocks AJAX request at the browser level so you can run front ends without really starting a backend server.",
+  "license": "MIT",
+  "homepage": "https://github.com/code-mancers/interceptor#readme",
+  "repository": {
```
revathskumar: I repository is already there and no need for change

package.json
```diff
@@ -1,4 +1,21 @@
 {
+  "name": "interceptor",
+  "version": "0.2.1",
+  "description":
+    "A browser extension that mocks AJAX request at the browser level so you can run front ends without really starting a backend server.",
```
revathskumar: run **frontend**

package.json
amit-mb: i just moved the same thing up. You can notice the deleted lines below  😄

package.json
amit-mb: 👍

package.json
revathskumar: in url `.git` is missing. 
Ideally this should be the url to `git` repo not its code home page.

PR #137, opened 2018-07-03
pretty print json on clicking fetch data button

amit-mb:


app/store/aliases.ts
```diff
@@ -29,7 +29,9 @@ const fetchDataAlias = (payload: payload) => {
       requestHeadersObject
     })
       .then((data: axios.AxiosResponse) => {
-        const stringifiedData = JSON.stringify(data.data);
+        const stringifiedData = data.headers["content-type"].includes("json")
```
revathskumar: Why we need this condition?

app/store/aliases.ts
```diff
@@ -29,7 +29,9 @@ const fetchDataAlias = (payload: payload) => {
...
+          ? JSON.stringify(data.data, null, 2)
+          : JSON.stringify(data.data);
```
revathskumar: Does this will create issue when the response is html/xml?

app/store/aliases.ts
amit-mb: we dont need the pretty printing if the `content-type` is anything other than `application/json`. I had earlier added a `strict` equals condition for checking, but turns out that some websites sent out `content-type` headers as `application/json; charset=utf-8`. So this condition is needed to filter out types like `text/html`/`text/plain` or `application/x-www-form-urlencoded` for instance

revathskumar: okey. but in the false part of this condition we are doing `JSON.stringify` doesn't that break for `text/html`?

amit-mb: This is quite strange. I logged out the following 

bug-json

and intercepted a svg file, for which the output is:

bug--json

running the following on chrome `console` or `node repl` gives me the following error

chrome-console

but we are getting the proper output for `stringified data` as you can see in first screenshot instead of it throwing an error. Not sure what is happening here.

amit-mb: This is what happens if I don't stringify the data for `svg` type (and what should happen ideally)

screen shot 2018-07-04 at 4 30 25 pm

revathskumar: I mean without `JSON.stringify` does the correct data coming to `textbox`?
like without escape?

amit-mb: yes, for non-json types, correct data is coming through. for `json` type, i am getting `[object Object]`

revathskumar: then why can't we remove the `JSON.stringify` for non JSON-types.

amit-mb: Done. Please check

PR #142, opened 2018-07-12
Setup lint staged

amit-mb:


package.json
```diff
@@ -36,7 +35,9 @@
     "enzyme-adapter-react-16": "^1.1.1",
     "enzyme-adapter-react-helper": "^1.2.3",
     "fork-ts-checker-webpack-plugin": "^0.4.1",
+    "husky": "^0.14.3",
     "jest": "^22.1.4",
+    "lint-staged": "^7.2.0",
```
revathskumar: Shouldn't we add `prettier` as `devDependencies`.

package.json
```diff
@@ -66,8 +72,21 @@
     "build": "webpack -p --config webpack.config.prod.js",
     "watch": "webpack --config webpack.config.dev.js --watch --progress --colors",
     "test": "jest",
-    "test:watch": "jest --watch"
+    "test:watch": "jest --watch",
+    "precommit": "lint-staged"
   },
+ "lint-staged": {
+   "linters":{
+       "*.{js,jsx,css}": [
```
revathskumar: Does prettier format css as well?

revathskumar: Do we have any `jsx` component?

package.json
```diff
@@ -66,8 +72,21 @@
...
+        "prettier --config .prettierrc --write",
+        "git add"
+      ],
+       "*.{ts,tsx}": ["prettier --config .prettierrc --parser typescript --write", "git add", "jest"]
+   },
+   "ignore": [
+    "/node_modules/"
+  ]
```
revathskumar: fix indendation

package.json
```diff
@@ -66,8 +72,21 @@
...
+},
```
revathskumar: Fix indendation

package.json
amit-mb: checked prettier, it supports `css` formatting with a plugin. removing it for now. Maybe we can use [stylelint](https://stylelint.io/) or [postcss](https://postcss.org/)

package.json
amit-mb: Do we need it? `prettier` is a devDependency for `lint-staged` and works for me out of the box (I dont have prettier installed globally as well)

amit-mb: However, `jsonlint` doesn't work out of box although that too is a `devDependency` for `lint-staged`. it asks me to do `npm install jsonlint`. Not sure whats going on

revathskumar: Looks like [prettier](https://github.com/okonet/lint-staged/blob/d52e074f66ae3c604413fca48f0ba6b999c2fcfb/package.json#L64) is available as dependency

package.json
```diff
@@ -66,7 +72,18 @@
...
+    "test:watch": "jest --watch",
+    "precommit": "lint-staged"
+  },
+  "lint-staged": {
+    "linters":{
+       "*.{js}": [
+        "prettier --config .prettierrc --write",
+        "git add"
+      ],
+       "*.{ts,tsx}": ["prettier --config .prettierrc --parser typescript --write", "git add", "jest"]
```
revathskumar: in `.prettierrc` we already specifying parser as typescript. So I think we don't need to specify this here

package.json
```diff
@@ -66,7 +72,18 @@
     "build": "webpack -p --config webpack.config.prod.js",
     "watch": "webpack --config webpack.config.dev.js --watch --progress --colors",
     "test": "jest",
-    "test:watch": "jest --watch"
+    "test:watch": "jest --watch",
+    "precommit": "lint-staged"
+  },
+  "lint-staged": {
+    "linters":{
+       "*.{js}": [
```
revathskumar: Please club this along with `ts` & `tsx`. One rule will be enough

PR #144, opened 2018-07-25
Remove fetch response icon and replace it with text

amit-mb:
Earlier - 
Earlier

Now -
Now


revathskumar:
can we use the text `Fetch Response`.
otherwise the UI is breaking.

amit-mb:
Is it better to just use `Server Response` ? `Fetch Response` seems vague as it doesn't specify which response.

As it is a button above the `textarea`, i think the users will be able to assume that it fills the textarea with `Server Response`

revathskumar:
or can the label be `Mock Reponse`? Either way we need to fix the UI.

revathskumar:
@amit-mb can you send the latest screenshot?

amit-mb:
@revathskumar this is the screenshot after making the latest commit

Fetch Response

PR #145, opened 2018-07-26
Display dynamic tooltip instead of static text for toggling Interceptor

amit-mb:
Earlier

Earlier

Now
Enable
Disable


revathskumar:
@amit-mb Why this changes? 
I personally in favor for label. 

Can have the current state `Enabled` or `Disabled` in the tooltip.

amit-mb:
I felt we can reduce the space in the top where it says `Toggle Interception` and user also gets dynamic text in popup. We can still keep the label BTW.

Not the current state but it is the other way around. 

When Interceptor is `ON` -> `Disable Interception`
When Interceptor is `OFF` -> `Enable Interception`

revathskumar:
Removing the label reducing the clarity and I don't think we need more space there right now.

PR #148, opened 2018-08-06
Add rule : Separate reducers

amit-mb:
- The reducer used earlier was used for all kind of actions. We separated the reducer into `rootReducer` and `addRequestReducer`

### DIFF
--- a/app/actions/addRequest.ts
+++ b/app/actions/addRequest.ts
@@ -1,0 +1,16 @@
+export const RESET = "RESET";
+export const UPDATE_REQUEST_FIELDS = "UPDATE_REQUEST_FIELDS";
+export const UPDATE_FIELD = "UPDATE_FIELD";
+
+export const updateAddRequestFields = (name, value) => {
+  return {
+    type: UPDATE_REQUEST_FIELDS,
+    payload: { name, value }
+  };
+};
+
+export const resetAddRequest = () => {
+  return {
+    type: RESET
+  };
+};
--- a/app/actions/addRequest.ts
+++ b/app/actions/addRequest.ts
@@ -1,11 +1,10 @@
 export const RESET = "RESET";
 export const UPDATE_REQUEST_FIELDS = "UPDATE_REQUEST_FIELDS";
-export const UPDATE_FIELD = "UPDATE_FIELD";
 
-export const updateAddRequestFields = (name, value) => {
+export const updateAddRequestFields = (url: string, method: string, error: string) => {
   return {
     type: UPDATE_REQUEST_FIELDS,
-    payload: { name, value }
+    payload: { url, method, error }
   };
 };
 
--- a/app/components/AddRuleModal.tsx
+++ b/app/components/AddRuleModal.tsx
@@ -1,8 +1,6 @@
 import * as React from "react";
-import { connect } from "react-redux";
-
+import * as uuid from "uuid";
 import { Modal } from "./ModalWrapper";
-import { updateAddRequestFields } from "./../actions/addRequest";
 
 interface AddRuleModalProps {
   addRequestDetails: Object;
@@ -11,7 +9,7 @@ interface AddRuleModalProps {
   tabId: number;
   updateRequest: (tabId: number, request: chrome.webRequest.WebRequestBodyDetails) => void;
 }
-class AddRuleModal extends React.PureComponent<AddRuleModalProps, {}> {
+export default class AddRuleModal extends React.PureComponent<AddRuleModalProps, {}> {
   isUrl = (str: string) => {
     try {
       new URL(str);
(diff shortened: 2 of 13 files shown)


package.json
```diff
@@ -58,6 +58,7 @@
       "js",
       "jsx"
     ],
+    "testURL": "http://localhost",
```
revathskumar: Why do we need this?

app/components/AddRuleModal.tsx
```diff
@@ -24,11 +21,9 @@ export class AddRuleModal extends React.PureComponent<AddRuleModalProps, {}> {
 
   handleClose = () => {
     const { props } = this;
+    //erase the previously set error message on each re-render
     //reset the url to empty string and request method to "GET"
-    props.updateAddRequestUrl("", props.tabId);
-    props.updateAddRequestMethod("GET", props.tabId);
-    //Erase the previously set error on opening the modal window
-    props.addRuleErrorNotify("", props.tabId);
+    props.updateAddRequestFields("", "GET", "");
```
revathskumar: This should happen on component unmount.

app/actions/addRequest.ts
```diff
@@ -0,0 +1,15 @@
+export const RESET = "RESET";
+export const UPDATE_REQUEST_FIELDS = "UPDATE_REQUEST_FIELDS";
+
+export const updateAddRequestFields = (url: string, method: string, error: string) => {
```
revathskumar: This method should accept only one param which is an object. that how we can generalize this method and use instead of n number of functions.

app/components/AddRuleModal.tsx
```diff
@@ -1,18 +1,15 @@
...
+import * as uuid from "uuid";
 import { Modal } from "./ModalWrapper";
 
 interface AddRuleModalProps {
-  addRequest: (url: string, method: string) => void;
-  addRequestMethod: string;
-  addRequestUrl: string;
-  addRuleErrorNotify: (errorMessage: string, tabId: number) => void;
-  addRuleError: string;
+  addRequestDetails: Object;
```
revathskumar: why this is mentioned as just Object. shouldn't we point this to interface of newRequest?

app/components/AddRuleModal.tsx
```diff
@@ -39,24 +34,36 @@ export class AddRuleModal extends React.PureComponent<AddRuleModalProps, {}> {
...
+        tabId: props.tabId,
+        type: "xmlhttprequest",
+        url: props.addRequestDetails.fields.url
+      };
+      props.updateRequest(props.tabId, requestObject);
       this.handleClose();
       return;
     } else {
-      props.addRuleErrorNotify("Please Enter a valid URL", props.tabId);
+      props.updateAddRequestFields(
```
revathskumar: here this should call like 
```js
props.updateAddRequestFields({error: "Please Enter a valid URL"}); 
```

app/components/AddRuleModal.tsx
```diff
@@ -65,10 +72,14 @@ export class AddRuleModal extends React.PureComponent<AddRuleModalProps, {}> {
               className="form-control"
               type="text"
               name="request_url"
-              defaultValue={props.addRequestUrl}
+              defaultValue={props.addRequestDetails.fields.url}
               id="url-input-modal"
               onChange={e => {
-                props.updateAddRequestUrl(e.target.value, props.tabId);
+                props.updateAddRequestFields(
```
revathskumar: similar to the comment before. should pass only the param needed.

app/components/AddRuleModal.tsx
```diff
@@ -77,9 +88,11 @@ export class AddRuleModal extends React.PureComponent<AddRuleModalProps, {}> {
             <select
               name="request_method"
               id="modal-request-method"
-              value={props.addRequestMethod}
+              value={props.addRequestDetails.fields.method}
               className="modal-method form-control"
-              onChange={e => props.updateAddRequestMethod(e.target.value, props.tabId)}
+              onChange={e =>
```
revathskumar: same issue here.

app/components/AddRuleModal.tsx
revathskumar: `AddRule` component can have instance method called `updateField`

```js
class AddRule extends Component {
  updateField = (e) => {
    const { name, value } = e.target;
    this.props.updateAddRequestFields({ [name]: value });
  }

  render(){
    return <input onChange={this.updateField} />
  }
}
```

app/reducers/rootReducer.ts
```diff
@@ -116,28 +113,25 @@ export const reducer = (state = INITIAL_POPUP_STATE, action: Action) => {
...
       };
-    case actionType.HANDLE_MODAL_METHOD_CHANGE:
-      return extendTabRecords(state, action.payload, {
-        addRequestMethod: action.payload.value
-      });
-    case actionType.HANDLE_MODAL_URL_CHANGE:
-      return extendTabRecords(state, action.payload, {
-        addRequestUrl: action.payload.value
-      });
+    // case actionType.HANDLE_MODAL_METHOD_CHANGE:
```
revathskumar: delete the unwanted code

revathskumar:
* [x] Rule Modal
* [x] Update count on adding url
* [x] Able to fetch response if the url is added via Rule modal
* [ ] Intercept the url added via Rule Modal

PR #150, opened 2018-08-20
#149 Disable interception on clearing requests

### LINKED ISSUE #149: After clear : the interception is still active
### Steps to reproduce
1. Record or add rule
2. Update the response and intercept
3. make the request and receive intercepted response
4. Clear the requests
5. make the request again and see the webpage receive intercepted response

### Expected
After clearing the requests the interception should not be active

### Actual
After clearing the requests the interception is still active

amit-mb:
Fixes #149

### DIFF
--- a/app/__mocks__/message_service.ts
+++ b/app/__mocks__/message_service.ts
@@ -1,10 +1,11 @@
-export const getRequests = jest.fn((tabId:number, callback) => {
+export const getRequests = jest.fn((tabId: number, callback) => {
   callback();
 });
 
+export const disableInterceptor = jest.fn();
 export const enableLogging = jest.fn();
 export const clearData = jest.fn();
 export const disableLogging = jest.fn();
-export const getEnabledStatus = jest.fn((tabId:number, callback) => {
-  callback()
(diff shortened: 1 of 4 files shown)


amit-mb:
No, this will not disable interception. To be precise, it will only trigger 
 ```js
      (function(){
        if(window.interceptor){
          window.interceptor.server.restore()
        }
      })();
```
You can test it once.

revathskumar:
@amit-mb But it gives the message `Interception Disabled` when we clear the list. 
The best way to fix this is not using the `restore()` instead do a new interception with empty list.

amit-mb:
@revathskumar the "Interception Disabled" message is due to this line in `content.ts` file 

https://github.com/code-mancers/interceptor/blob/b4392ddb1a23c8338a25bb7fc97c1ede4e6c0543/app/content/content.ts#L101

We can show this message only when the user toggles to disabled state. 

But we will have to do a `restore()` to clear the current interceptions done by `sinon`. Is there any other way to do a new interception without doing a `restore` ?

revathskumar:
@amit-mb Let me check and come back.

PR #152, opened 2018-08-21
fix(popup): clear active interception when clear recorded data

### LINKED ISSUE #149: After clear : the interception is still active

revathskumar:
Fix #149

### DIFF
--- a/app/message_service.ts
+++ b/app/message_service.ts
@@ -11,12 +11,12 @@ export function disableLogging(tabId: number) {
 
 export function clearData(tabId: number) {
   chrome.runtime.sendMessage({ message: "CLEAR_DATA", tabId });
+  chrome.tabs.sendMessage(tabId, { message: "CLEAR_DATA" });
 }
 
-export function interceptChecked(tabId: number, requests: Array<any>) {
+export function interceptChecked(tabId: number) {
   chrome.tabs.sendMessage(tabId, {
     message: "INTERCEPT_CHECKED",
-    requestsToIntercept: requests,
     tabId
   });
 }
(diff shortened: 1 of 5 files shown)


amit-mb:
Yes, this works without having to call `server.restore()`.

In addition to this. We need to clear the store values as well.

Attaching a screenshot of store values after clearing the store: 
screen shot 2018-08-21 at 2 23 58 pm

We need to clear `requests` and `requestRecords`. Shall I take this up? @revathskumar

app/message_service.ts
```diff
@@ -11,12 +11,12 @@ export function disableLogging(tabId: number) {
 
 export function clearData(tabId: number) {
   chrome.runtime.sendMessage({ message: "CLEAR_DATA", tabId });
```
revathskumar: @amit-mb Is message is suppose to clear the data from redux store right?

revathskumar:
@amit-mb Please test it again.

app/reducers/rootReducer.ts
```diff
@@ -126,7 +126,8 @@ export const reducer = (state = INITIAL_POPUP_STATE, action: Action) => {
       });
     case actionType.CLEAR_REQUESTS:
       return extendTabRecords(state, action.payload, {
-        requests: []
+        requests: [],
+        requestRecords: {}
```
revathskumar: @amit-mb  I think now the requestRecords will get cleared from the redux store.

amit-mb: Yes, this is what I was referring to 😄

PR #159, opened 2018-08-24
perf: loads fonts from local

amit-mb:
- loads fonts from local instead of, from the cdn


webpack.config.dev.js
```diff
@@ -48,11 +48,12 @@ module.exports = {
...
-      {from: "manifest.json"},
-      {from: "index.html"},
-      {from: "./lib/*"},
-      {from: "stylesheets/*"},
-      {from: "images/*"}
+      { from: "manifest.json" },
+      { from: "index.html" },
+      { from: "./lib/*" },
+      { from: "stylesheets/*" },
+      { from: "stylesheets/fonts/*" },
```
aksh93: @amitmbee is this being redundant after defining `{ from: "stylesheets/*" }` ?

aksh93:
@amitmbee Please do update with the latest master

Issue #164, opened 2018-11-08
text/javascript not in Content-Type dropdown


mireczatko:
text/javascript not in Content-Type dropdown, there is only new application/javascript

PR #165, opened 2019-01-12
feat: add support for text/javascript content-type

amit-mb:
#164


revathskumar:
@amit-mb if the content-type is `text/javascript`. Does it get executed? 
Isn't this will be security threat. if we allow to execute js

